PR #1, opened 2022-08-30
Packaging updates and minor tinkering

LiamPattinson:
I updated the setuptools configuration to use the newer declarative config method using `setup.cfg` and `pyproject.toml` (see [here](https://setuptools.pypa.io/en/latest/userguide/declarative_config.html), [PEP 517](https://peps.python.org/pep-0517/) and [PEP 518](https://peps.python.org/pep-0518/). Annoyingly, they're  in the process of replacing this system by an even newer one that puts everything in `pyproject.toml`, but that's probably about a year away. This makes it easier to `pip install`, as it will now automatically grab dependencies of the project, including the build system itself.

This also makes it trivial to upload to PyPI. You can just do the following:

```bash
pip install build twine
python -m build # bundles package, puts a .whl and .tar.gz in dist/
twine check dist/* # check everything is good to go
twine upload dist/* # Send it to PyPI. Can now pip install from anywhere!
```
I tested it out by pushing it to [TestPyPI](https://packaging.python.org/en/latest/guides/using-testpypi/), and it [works just fine](https://test.pypi.org/project/archeryutils/0.1.1/). You can try it out with:

```bash
pip install --index-url https://test.pypi.org/simple archeryutils
```
I also moved modules into a `archeryutils` directory, so there's a single entrypoint to the library. The dir `archerycls` was removed, and the modules inside are now available at the top level of `archeryutils`, and `handicaps` is now a subpackage of `archeryutils`. `handicap_equations` and `handicap_functions` can be imported using `import archeryutils.handicap_equations` etc. (no need to import via `handicaps`).

Other random updates:
- Some simplification at the end of `rounds.py` when importing all the rounds jsons.
- Simplified `print_handicap_table`: now creates a single string that can be used either when writing to terminal or writing to
file -- just reduces some repeat code.
- black formatting (I realise it's doing something awful to the `arrow_score` equations -- we may need to look at a better way to
achieve those! Maybe it could be a function of `Target`?)
- Updates to keep flake8 happy (shortened long strings/comments, removed unused imports, etc)

Let me know what you think!


archeryutils/handicaps/handicap_equations.py
```diff
@@ -244,7 +248,9 @@ def arrow_score(target, h, hc_sys, hc_dat, arw_d=None):
 
     if target.scoring_system == "5_zone":
         s_bar = (
-            9 - 2 * sum(
+            9
+            - 2
+            * sum(
```
jatkinson1000, line 253: I am willing to argue with the Black formatting here. I think everything on a single line is less readable. Is there a good argument for changing it as it suggests?

archeryutils/handicaps/handicap_equations.py
```diff
@@ -12,8 +12,6 @@
 import json
 from dataclasses import dataclass
 
-from archerycls import targets, rounds
```
jatkinson1000, line 15: If removing this line do the docstrings need changing to find the correct types for Target, Round, and Pass?

archeryutils/handicaps/handicap_functions.py
```diff
@@ -158,7 +115,37 @@ def print_handicap_table(
         )
         print("Done.")
 
-    return None
```
jatkinson1000, line 161: Isn't it good practice to have a return None at the end of the function (print_handicap_table)?

jatkinson1000 (commented):
All looks sensible to me in general. Left a couple of comments about things that could change.
Also I note you didn't update the file headers, however, the 'last modified by' seems superfluous and is a hangover habit from before I used git properly, so I can probably remove them.

archeryutils/handicaps/handicap_equations.py
LiamPattinson: It's an issue I've seen a lot with scientific code, since any serious equations can very easily run over the 88 characters max line length. I think the reasoning behind black here is that it's more consistent: each line is a self-contained expression, and operators are used as the line breaks. I agree that it looks awful though... I just left it the way black says to do it, since otherwise I'd have to go back and manually fix it every time I wanted to check my formatting. I think it might be possible to generalise the equations in some way so that we don't need the `if...elif...else` block, and we'll be able to write them cleaner in whatever new form they take. Something like:

```python
s_bar = target.max_score
# handle innermost ring
s_bar -= np.exp(-((target.inner_diameter + arw_rad) / sig_r) ** 2)
# handle other rings
s_bar -= target.score_increment * (
    np.sum(
        np.exp(
            -(((np.arange(2, tar_rings) * tar_dia / tar_zones) + arw_rad)  / sig_r) ** 2
        )
    )
)
# handle outermost ring
#...
# handle missing rings
# etc... probably got an unmatched parenthesis in there
```
I think this would be a better approach overall, as otherwise the handicap system needs to know how to interpret "10 zone 5 ring compound" instead of just using a generic Target. Having a large collection of similar equations is also quite error prone compared to having a single generic method.

archeryutils/handicaps/handicap_equations.py
LiamPattinson: I'm honestly not sure. One good excuse to keep them would be to add type hinting to the function itself. I'll get on that in a minute...

archeryutils/handicaps/handicap_functions.py
LiamPattinson: I think it might just be personal preference. I usually exclude it, since you don't tend to return from `void` functions in C/C++, and I picked up most of my bad habits there. I don't think it changes behaviour either way. I guess that it might help the reader to see when one function ends and another begins?

archeryutils/handicaps/handicap_equations.py
jatkinson1000: OK, yes, I like this idea.
Requires some sitting down with a pen and paper for a while...
But could be made general based on intrinsic properties of Target.
Not something to be done on this commit, however, so I'll open an issue.
And fine - given it doesn't need to be looked at too often I am happy to go with the black. ;P

archeryutils/handicaps/handicap_functions.py
jatkinson1000: Yes, I prefer it, so if you can leave it in that would be great, thanks.

LiamPattinson: I added it back in before I read your comment, but as a plain `return` instead of `return None`. I'd argue there's a difference in intent between the two: plain `return` says "the function exits here", while `return None` says "I'm explicitly returning a None which should be handled elsewhere in the code". Up to you though.

LiamPattinson:
I undid some of the black formatting and added type hints. Unlike the docstrings, some IDEs and static analysis tools can use type hints to verify that the correct types are being passed around, and libraries like Pydantic can use them to enforce stricter typing, but otherwise they're just annotations. It took me a little while to come around to them, but nowadays I quite like them as they tell you what a function argument is supposed to be at a glance. (It also means you can continue importing relevant libraries without flake8 getting all shouty.)

jatkinson1000 (approved):
All looks good.

PR #5, opened 2022-12-21
Rootfinding

jatkinson1000:
Write own rootfinding algorithm to remove single small dependency on scipy.


jatkinson1000:
Looks good ;)

PR #8, opened 2022-12-22
Fixes for pyproject.toml install issues

LiamPattinson:
I don't normally use `pyproject.toml`-only builds, so I have no idea what I'm doing.


jatkinson1000:
Just pulled and it seems to be working OK with a pip import via git!

PR #9, opened 2022-12-22
Adding __init__.py file to round_data_files folder

DMcP89:
This will let setup tools pick up this folder while installing the package

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -27,6 +27,10 @@ dependencies = [
     "numpy>=1.20.3",
 ]
 
+packages = [
+    {include = "archeryutils"},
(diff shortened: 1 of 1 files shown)


jatkinson1000:
Thank you for the contribution!
A separate pull request came in ahead of you by about 10 mins that dealt with this by adding the json data files to the pyproject.toml file.
This all seems to work so I merged that and will close this.

If I understand correctly including the __init__.py would make those files accessible from outside of the module once installed?

DMcP89:
It should yes. But I believe including them in the pyproject.toml will as well. Definitely the better solution.

Issue #23, opened 2023-03-18
Tests needed for classifications


jatkinson1000:
Tests for handicaps have been added, but are yet to be written for classifications.

Issue #3, opened 2022-08-30
Simplify importing classes/functions


LiamPattinson:
I think it'd be a good idea to expose some classes/functions to the top-level namespace. It'd be nice if we could do stuff like:

```python
import archeryutils as au
my_target = au.Target(...)
```
We'll have to be fairly selective though, as most of the internals probably don't need to be part of the user API.

jatkinson1000:
Yep, probably easiest to arrange a meeting to discuss this at a convenient point?

jatkinson1000:
@LiamPattinson I have implemented this over in #55 if you want to take a look and share any thoughts/review.

PR #42, opened 2023-08-20
Classifications refactor

### LINKED ISSUE #39: Refactor classifications
Classifications has grown to become rather behemothic.
Consider refactoring into separate files for each classification.

jatkinson1000:
Closes #39 
~Consider also tackling #24~

Refactor excessively long classifications file into several separate ones for each type of classification (plus general utils file).

It would be nice for the code to continue functioning as it previously did with
```
import archeryutils.classifications as class_func
```
rather than having to import each type of classification (indoor, outdoor, etc.) separately.
I have made an effort to do this using the classifications module `__init__.py` but it's not quite working.
@LiamPattinson is there any chance you could cast an eye and advise me what is best to do?
This could also be related to issue #3 

TODO:
- [ ] Update examples notebook


LiamPattinson:
Sorry, I'm not really sure what behaviour you're trying to retain. Please could you post example snippets of how you'd like things to work and how things have to work instead due to this PR?

LiamPattinson:
I think I've figured it out. In your top-level `__init__.py`, you have:

```python
from archeryutils.classifications import classifications
```

So when you import `archeryutils.classifications`, you're actually importing the module `archeryutils.classifications.classifications`, which is now an empty module. It should work if you change the top-level `__init__.py` to instead say:

```python
import archeryutils.classifications as classifications
```

jatkinson1000:
> I think I've figured it out. Change the top-level `__init__.py` to instead say:
> 
> ```python
> import archeryutils.classifications as classifications
> ```

M8 that's done it.
Thank you so much, pypack guru.

jatkinson1000:
Adding classification tests so that this PR will also close #23 
This is necessary here to ensure everything is working OK after a major refactor.

Issue #28, opened 2023-03-21
Get CI green (address pylint)


jatkinson1000:
Currently the CI checks fail for pylint (added in #25 ).
Whilst some of these are legitimate to ignore (variable names) others hint at the need for some refactoring and will be addressed in some other issues.

After looking at these we should then return to place any ignores that we feel are legitimately warranted.

Issues to tackle that will help address this:
- #24 
- #2

jatkinson1000:
Short variable name warnings will be dealt with in an upcoming release of PyLint: https://github.com/pylint-dev/pylint/pull/8813

PR #44, opened 2023-09-02
Linting

jatkinson1000:
Addresses #28 

Clean up code to address pylint errors, in some cases leading to notable refactors => merge as rebase.


jatkinson1000:
Refactor of handicap params leads to failure of tests due to type hinting not being fully available in python <3.9
See [here](https://peps.python.org/pep-0585/#backwards-compatibility).

Since 3.7 has reached end of life, and 3.8 has 1 year left this could perhaps be solved by updating rthe supported versions for this library to 3.9+

PR #51, opened 2024-01-09
Separate score for round and score for passes into two functions

### LINKED ISSUE #4: hc_eq.score_for_round should not return a tuple
Currently hc_eq.score_for_round returns a tuple of the score for the round and the score for each distance.
This is not an intuitive use, however.

Need a function hc_eq.score_for_round that returns just a score.
Need a new function hc_eq.score_for_round_distances that returns an array.

Need to rectify this soon as it will have a lot of knock on effects elsewhere such as classifications and generating handicap tables.

jatkinson1000:
Closes #4 

Add `score_for_passes` function that can be called by `score_for_round` such that the latter now returns a single number which is more intuitive.

This is a Breaking Change implemented in preparation for first release.

### DIFF
--- a/archeryutils/classifications/agb_indoor_classifications.py
+++ b/archeryutils/classifications/agb_indoor_classifications.py
@@ -239,7 +239,7 @@ def agb_indoor_classification_scores(
             hc_scheme,
             hc_params,
             round_score_up=True,
-        )[0]
+        )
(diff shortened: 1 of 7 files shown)


jatkinson1000:
Codecov diff is complaining about a line that wasn't hit in previous testing, so OK to merge.

PR #52, opened 2024-01-10
Targetdiameter

### LINKED ISSUE #27: Units for target face diameter
### Discussed in https://github.com/jatkinson1000/archeryutils/discussions/17

- Add option for user to specify target face diameter unit
- Set default to be [cm]
- Store under the hood as [m] to preserve SI and not affect code elsewhere

This requires:
- Add optional argument to `Target` class
- Add conversion code as part of `__init__` in `Target` class
- Add `INCHES2M` and `CM2M` to `constants` file
- Add tests for conversion process
- Update and round data files accordingly

TomHall2020:
Closes #27 
Set default target diameter units to cm
Updates tests and load_rounds functionality
Round data files all seem to specify target face sizes in cm already so enforcing this as the default unit makes sense.

Aliases for distance units moved into the constants file as it was getting out of hand and avoiding repitition, choice of conversion constants handled in the branching points in `Target.__init__`, this could perhaps be handled more elegantly if more units need to be supported in the future but didn't want to change more than necessary for now.

### DIFF
--- a/archeryutils/constants.py
+++ b/archeryutils/constants.py
@@ -1,14 +1,13 @@
 """Constants used in the archeryutils package."""
-from types import SimpleNamespace
 
-TO_METRES = {
+_CONVERSIONS = {
     "metre": 1.0,
     "yard": 0.9144,
     "cm": 0.01,
     "inch": 0.0254,
 }
 
-YARD_ALIASES = {
+_YARD_ALIASES = {
     "Yard",
     "yard",
     "Yards",
@@ -21,7 +20,7 @@ YARD_ALIASES = {
     "yds",
 }
 
-METRE_ALIASES = {
+_METRE_ALIASES = {
     "Metre",
     "metre",
     "Metres",
@@ -32,7 +31,7 @@ METRE_ALIASES = {
     "ms",
 }
 
-CM_ALIASES = {
+_CM_ALIASES = {
     "Centimeter",
     "centimeter",
     "Centimeters",
@@ -43,29 +42,46 @@ CM_ALIASES = {
     "cms",
 }
 
-INCH_ALIASES = {
+_INCH_ALIASES = {
     "Inch",
     "inch",
     "Inches",
     "inches",
 }
 
-
-DistanceUnits = SimpleNamespace(
-    yard = YARD_ALIASES,
-    metre = METRE_ALIASES,
-    cm = CM_ALIASES,
-    inch = INCH_ALIASES,
+_ALIASES = dict(
+    yard = _YARD_ALIASES,
+    metre = _METRE_ALIASES,
+    cm = _CM_ALIASES,
+    inch = _INCH_ALIASES,
 )
 
-def normalise_unit_name(unit: str) -> str | None:
-    """Convert any supported unit name alias into a cannonical string representation"""
-    if unit in YARD_ALIASES:
-        return 'yard'
-    if unit in METRE_ALIASES:
-        return 'metre'
-    if unit in CM_ALIASES:
-        return 'cm'
-    if unit in INCH_ALIASES:
-        return 'inch'
-    return None
+class Length:
+    yard = _YARD_ALIASES
+    metre = _METRE_ALIASES
+    cm = _CM_ALIASES
+    inch = _INCH_ALIASES
+
+    _reversed = {
+        alias: name
+        for name in _CONVERSIONS
+        for alias in _ALIASES[name]
+    }
+
+    _conversions = {
+        alias: factor
+        for name, factor in _CONVERSIONS.items()
+        for alias in _ALIASES[name]
+    }
+
+    @classmethod
+    def to_metres(cls, value: float, unit: str) -> float:
+        return cls._conversions[unit] * value
+
+    @classmethod
+    def from_metres(cls, metre_value: float, unit: str) -> float:
+        return metre_value / cls._conversions[unit]
+
+    @classmethod
+    def definitive_name(cls, alias: str) -> str:
+        return cls._reversed[alias]
--- a/archeryutils/constants.py
+++ b/archeryutils/constants.py
@@ -40,7 +40,8 @@ CM_ALIASES = (
 )
 
 
-class DistanceUnits(SimpleNamespace):
-    yard = YARD_ALIASES
-    metre = METRE_ALIASES
-    cm = CM_ALIASES
+DistanceUnits = SimpleNamespace(
+    yard = YARD_ALIASES,
+    metre = METRE_ALIASES,
+    cm = CM_ALIASES,
+)
--- a/archeryutils/constants.py
+++ b/archeryutils/constants.py
@@ -1,11 +1,14 @@
 """Constants used in the archeryutils package."""
 from types import SimpleNamespace
 
-YARD_TO_METRE = 0.9144
-CM_TO_METRE = 0.01
-INCH_TO_METRE = 0.0254
+TO_METRES = {
+    "metre": 1.0,
+    "yard": 0.9144,
+    "cm": 0.01,
+    "inch": 0.0254,
+}
 
-YARD_ALIASES = (
+YARD_ALIASES = {
     "Yard",
     "yard",
     "Yards",
@@ -16,9 +19,9 @@ YARD_ALIASES = (
     "yd",
     "Yds",
     "yds",
-)
+}
 
-METRE_ALIASES = (
+METRE_ALIASES = {
     "Metre",
     "metre",
     "Metres",
@@ -27,9 +30,9 @@ METRE_ALIASES = (
     "m",
     "Ms",
     "ms",
-)
+}
 
-CM_ALIASES = (
+CM_ALIASES = {
     "Centimeter",
     "centimeter",
     "Centimeters",
@@ -38,14 +41,14 @@ CM_ALIASES = (
     "cm",
     "CMs",
     "cms",
-)
+}
 
-INCH_ALIASES = (
+INCH_ALIASES = {
     "Inch",
     "inch",
     "Inches",
     "inches",
-)
+}
 
 
 DistanceUnits = SimpleNamespace(
@@ -54,3 +57,15 @@ DistanceUnits = SimpleNamespace(
     cm = CM_ALIASES,
     inch = INCH_ALIASES,
 )
+
+def normalise_unit_name(unit: str) -> str | None:
+    """Convert any supported unit name alias into a cannonical string representation"""
+    if unit in YARD_ALIASES:
+        return 'yard'
+    if unit in METRE_ALIASES:
+        return 'metre'
+    if unit in CM_ALIASES:
+        return 'cm'
+    if unit in INCH_ALIASES:
+        return 'inch'
+    return None
--- a/archeryutils/constants.py
+++ b/archeryutils/constants.py
@@ -50,11 +50,12 @@ _INCH_ALIASES = {
 }
 
 _ALIASES = {
-    "yard" : _YARD_ALIASES,
-    "metre" : _METRE_ALIASES,
-    "cm" : _CM_ALIASES,
-    "inch" : _INCH_ALIASES,
+    "yard": _YARD_ALIASES,
+    "metre": _METRE_ALIASES,
+    "cm": _CM_ALIASES,
+    "inch": _INCH_ALIASES,
 }
+
 
 class Length:
     """
@@ -86,11 +87,7 @@ class Length:
     cm = _CM_ALIASES
     inch = _INCH_ALIASES
 
-    _reversed = {
-        alias: name
-        for name in _CONVERSIONS_TO_M
-        for alias in _ALIASES[name]
-    }
+    _reversed = {alias: name for name in _CONVERSIONS_TO_M for alias in _ALIASES[name]}
 
     _conversions = {
         alias: factor
--- a/archeryutils/constants.py
+++ b/archeryutils/constants.py
@@ -49,14 +49,38 @@ _INCH_ALIASES = {
     "inches",
 }
 
-_ALIASES = dict(
-    yard = _YARD_ALIASES,
-    metre = _METRE_ALIASES,
-    cm = _CM_ALIASES,
-    inch = _INCH_ALIASES,
-)
+_ALIASES = {
+    "yard" : _YARD_ALIASES,
+    "metre" : _METRE_ALIASES,
+    "cm" : _CM_ALIASES,
+    "inch" : _INCH_ALIASES,
+}
 
 class Length:
+    """
+    Utility class for Length unit conversions
+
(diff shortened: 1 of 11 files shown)


TomHall2020:
Just saw your message on the issue after posting this, haven't added the units option to Pass yet but would be trivial to add if you want it still.

jatkinson1000 (changes requested):
Thanks for starting this @TomHall2020, and thanks for getting involved with the project. Progress looks good so far.
I particularly like the condensing of the possible units terms which makes things much cleaner!

I have picked out a few specific comments in the code below, but more generally:
- I would like to have inches as an option (Worcester and IFFA etc.) if you could add this.
  - A knock on from this is that it would probably be a good idea to update the round loader to accept a non-default diameter unit. I could look at this later however if you want.
- it should be possible to pass distance units to the `Pass` constructor as an optional argument (as you already mention in a comment above).

I'm getting a few spurious additional blank lines and re-arranged imports etc. Are you using an 'intelligent' editor/IDE that has automatically done this?

archeryutils/targets.py
```diff
@@ -60,41 +63,32 @@ def __init__(
...
+                f"Diamter unit '{diameter_unit} not recognised. "
+                "Select from 'cm' or 'metre'"
+            )
+
         self.native_dist_unit = native_dist_unit
-        self.distance = (
-            distance * YARD_TO_METRE if self.native_dist_unit == "yard" else distance
-        )
+        self.distance = distance
+        self.diameter_unit = diameter_unit
```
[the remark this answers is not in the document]
TomHall2020: Implemented but it gets quite verbose in the constructor:
Also a little confusing with the diameter being the first positional argument and then the units coming at the end, when writing out you would think to specify them together.

`Target(16, "Worcester", 20, "yards", indoor=True, native_diameter_unit="inches")`

Perhaps out of scope but I think for the same character count it would be better to have everything keyword and terser:

`Target(diam=16, diam_unit="inches", dist=20, dist_unit="yards", indoor=True)`

Or batch together lengths and units into a single object and pass those around instead, though thats more disruptive to the api:
```
#example length object, could implement this properly with typing etc
Length = namedtuple("Length", ["value", "unit"])
Target(diameter=Length(16, "inches"), distance=Length(20, "yards"), indoor=True)
```

archeryutils/tests/test_targets.py
```diff
@@ -34,29 +34,44 @@ def test_invalid_distance_unit(self) -> None:
...
+    def test_invalid_diameter_unit(self) -> None:
+        with pytest.raises(ValueError):
+            Target(122, "5_zone", 50, "yards", diameter_unit="InvalidDiamterUnit")
+
+    @pytest.mark.xfail(
+        reason="Only implemented cm conversions in #27",
+        raises=ValueError,
+    )
+    def test_unsupported_diameter_unit(self) -> None:
+        Target(16, "Worcester", 20, "yards", indoor=True, diameter_unit="inches")
```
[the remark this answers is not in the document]
TomHall2020: Removed and replaced by checking the ValueError. Was more in here to draw attention to specifically missing inches as the other commen unit which you've mentioned wanting to support anyway

archeryutils/targets.py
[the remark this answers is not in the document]
jatkinson1000: Mmmm yes - I would support removing `native_` from both as somewhat redundant now.
And I agree - when reading through your original changes I also felt it would be better to specify magnitude and unit together.
Batching together I don't think is ideal, as we want the units to be optional to simplify construction in general.

After some thought I think ordering arguments as `(scoring system, dia, dist, dia_unit, dist_unit...)` would be OK, as users can construct simply using assumed units, but if specifying as kwargs could group the unit with the distance as they please e.g. `("10_zone", dia=16, dia_unit="inches", dist=20, dist_unit="yards", indoor=True))` (I'd just need to check that definitely works). What do you think?

archeryutils/constants.py
```diff
@@ -1,2 +1,46 @@
...
+    "cm",
+    "CMs",
+    "cms",
+)
+
+
+class DistanceUnits(SimpleNamespace):
+    yard = YARD_ALIASES
+    metre = METRE_ALIASES
+    cm = CM_ALIASES
```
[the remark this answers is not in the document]
TomHall2020: Huh, I've always used the subclassing but never caught that complaint before. I'd argue with pylint that that is a perfectly reasonable way to use a namespace, dataclass seems wierd to me as we're never going to instantiate it with different values. But also can keep it quite by just instatitating the namespace instead which I've gone for here for simplicty.

archeryutils/tests/test_targets.py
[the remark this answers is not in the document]
jatkinson1000: Thanks. I can't see any more commits, so assume this is local you are referring to?

TomHall2020:
Updated with Inches support, passthrough of units and I think most of the comments addressed. 
I agree that the unit conversion constants are out of hand now. Plus the conversion logic occurs in multiple places now, at least the `Target` constructor and the `Round.get_info` method. 

I can clean that up either by 
1. namespacing into a dict or similar like suggested, and still doing the conversion logic at those call sites
2. factoring the conversion logic out into a seperate function

RE my idea on the call signature for targe and packaging lengths up into their own object, just beacuse if you did want to go down that road then here option 1 is best in this PR and doing that change makes most sense done seperately? Although appreicate its another breaking change to the signature.

Otherwise happy to do 1 or 2 and close this off.

archeryutils/targets.py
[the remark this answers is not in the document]
TomHall2020: Definitely agree with scoring system as first arg. 
Positional argument calls and relying on defaults can still be a bit fugly:

```
#not obvious which is distance and which is diameter!
Target("10_zone", 20, 40)
```

archeryutils/rounds.py
```diff
@@ -1,8 +1,8 @@
 """Classes to define a Pass and Round for archery applications."""
-from typing import List, Union, Tuple
+from typing import List, Tuple, Union
 
-from archeryutils.targets import Target
 from archeryutils.constants import YARD_TO_METRE
+from archeryutils.targets import Target
 
 
 class Pass:
```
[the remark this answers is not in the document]
TomHall2020: Done, could probably do with improved testing for it

TomHall2020:
Had another look at this, if you check the latest two commits. First implements namespacing the conversions into a dictionary and getting rid of individual constants `CM_TO_METRE` etc.

Second takes it a step further and packages everything together into a `Length` class that is built from the definited constants and contains functions to convert to/from metres and also de-alias unit names. In my opinion it makes the code elsewhere much more readable. If you like the idea then I can add unit tests specifically for that, otherwise can stick with the dictonary approach in the previous commit and just let the current test suite cover it.

archeryutils/rounds.py
```diff
@@ -160,11 +166,7 @@ def max_distance(self, unit: bool = False) -> Union[float, Tuple[float, str]]:
         """
         max_dist = 0.0
         for pass_i in self.passes:
-            dist = (
-                pass_i.distance / YARD_TO_METRE
-                if pass_i.native_dist_unit == "yard"
-                else pass_i.distance
-            )
+            dist = Length.from_metres(pass_i.distance, pass_i.native_dist_unit)
             if dist > max_dist:
```
TomHall2020, line 170: While checking the change here noticed a bug, max_distance is comparing across units. A round with mixed distance units will give the wrong result here (eg 75m > 80y). Not common, but will come up for anyone making custom rounds, or potentially in IFAA where feet and yards are mixed.

TomHall2020: Can include a test and fix in next commits, assuming this was undesired behaviour

jatkinson1000: Nice spot, yes this should give the longest distance irrespective of unit of measurement.
Fix appreciated.

archeryutils/targets.py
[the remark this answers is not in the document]
TomHall2020: What about batching but allowing either a single integer value with default units, or the tuple with units specified, and then dealing with it in the constructor? Then can have distance and diameter as keywords and leave the system as the first and only positional parameter

```
#implicit, default units
Target("10_zone", distance=90, diameter=122)
 
#explicit units, can mix and match with default units, order of distance vs diameter doesn't matter, still fairly concise
Target("10_zone", distance=(40, "metres"), diameter=40)
Target("Worcster", diameter=(16, "inches"), distance=(20, "yards"), indoors=True)
```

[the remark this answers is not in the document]
jatkinson1000: After thinking about this I think this sort of thing is probably the best approach.
With a professional hat on I'd probably say to implement in an alternative PR, but happy for it to be done here if you feel up to it.

archeryutils/constants.py
```diff
@@ -1,2 +1,87 @@
 """Constants used in the archeryutils package."""
-YARD_TO_METRE = 0.9144
+
+_CONVERSIONS = {
+    "metre": 1.0,
+    "yard": 0.9144,
+    "cm": 0.01,
+    "inch": 0.0254,
+}
```
jatkinson1000, line 8: I'd rename this to `_CONVERSIONS_TO_M` or similar, just to be explicit.

archeryutils/constants.py
```diff
@@ -1,2 +1,87 @@
...
+    def to_metres(cls, value: float, unit: str) -> float:
+        return cls._conversions[unit] * value
+
+    @classmethod
+    def from_metres(cls, metre_value: float, unit: str) -> float:
+        return metre_value / cls._conversions[unit]
+
+    @classmethod
+    def definitive_name(cls, alias: str) -> str:
+        return cls._reversed[alias]
```
jatkinson1000, line 87: ~Rather than 'definitive' perhaps consider 'native_units' or 'original_units'? - one could argue that the 'definitive' unit of distance in this code is metres since everything is converted to that before applying equations.~

After figuring out how this works I wonder if it is neater to preserve the original functionality and just force any 'alternative' input format to be standardised to 'metres', 'yards' etc. This would be more consistent throughout and simplify the logic.

I would still consider a rename to `definitive_units` rather than `definitive_name`.

archeryutils/targets.py
```diff
@@ -59,42 +63,24 @@ def __init__(
...
-            native_dist_unit = "metre"
-        else:
+        if native_dist_unit not in Length.yard | Length.metre:
             raise ValueError(
                 f"Distance unit '{native_dist_unit}' not recognised. "
                 "Select from 'yard' or 'metre'."
             )
+        distance = Length.to_metres(distance, native_dist_unit)
 
+        if native_diameter_unit not in Length.cm | Length.inch | Length.metre:
```
jatkinson1000, line 73: Please use logical `or` rather than bitwise `|`.

archeryutils/constants.py
```diff
@@ -1,2 +1,87 @@
...
+    "Inches",
+    "inches",
+}
+
+_ALIASES = dict(
+    yard = _YARD_ALIASES,
+    metre = _METRE_ALIASES,
+    cm = _CM_ALIASES,
+    inch = _INCH_ALIASES,
+)
```
jatkinson1000, line 57: For consistency with above (and preferred approach) use `{...}` rather than `dict(...)`.

archeryutils/constants.py
```diff
@@ -1,2 +1,87 @@
...
+
+class Length:
```
jatkinson1000, line 59: Took me a while to figure this out, so docstrings definitely required.
I have been following the [numpy conventions](https://numpydoc.readthedocs.io/en/latest/format.html#documenting-classes) for this code.

jatkinson1000 (changes requested):
Yep, like the idea of a `Length` class. Couple of thoughts on the finer implementation that I will add in-code comments for.

Also dropped a comment on the 'magnitude and units' conversation.

Finally - you may want to run [black](https://black.readthedocs.io/en/stable/index.html) on the code pre-submission.

archeryutils/targets.py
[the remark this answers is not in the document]
jatkinson1000: Such a change will also require updating the loader (and other things...?) which I am happy to look at if you want.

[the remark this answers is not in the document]
TomHall2020: I'm happy to keep it seperate, this one already expanded in scope quite a bit.

archeryutils/targets.py
TomHall2020: Means different things? Logical or would be ~slightly~ a lot more verbose:
```python
# have to check membership of each unit set seperately using logical ors
if native_diameter_unit not in Length.cm or native_diameter_unit not in Length.inch or native_diameter_unit not in Length.metre:
    ...

# bitwise or creates union of sets of aliases, equivalent to:
if native_diameter_unit not in set.union(Length.cm, Length.inch, Length.metre):
    ...

# parens for emphasis, this is definitely not what we want
if (native_diameter_unit not in Length.cm) or Length.inch or Length.metre
    # returns true if variable is in cm or otherwise returns the inch set
```

archeryutils/constants.py
TomHall2020: Yup thats fair, theres a bit of wrangling to get all the converisons and aliases into useful structures without repitition. Have drafted them up already will add with next push

archeryutils/targets.py
jatkinson1000: Huh. TIL: Certain operators are overloaded in python to perfom set logic.
So happy to leave as is. :)

In general bitwise logic operators are intended for integers only, and [lead to some gotchas](https://stackoverflow.com/questions/3845018/boolean-operators-vs-bitwise-operators) if applied inappropriately.

archeryutils/targets.py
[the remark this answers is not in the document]
jatkinson1000: Yep, OK, I'll open an issue and do it once this PR is merged.

archeryutils/constants.py
TomHall2020: `definitive_units` is a better method name I agree.

This PR still forces standardisation of any input to the definitve names, it just includes the aliases in the conversion dictionaries as well I just found that the coercion to a definitive name fitted more neatly factored into the length class rather than in the init functions where it was used.

jatkinson1000: Ah, yes, I misinterpreted what was happening with the `_reversed` variable and `definitive` method when reading previously. #docstrings

TomHall2020:
Think this should be pretty much there now.

jatkinson1000 (approved):
Thanks @TomHall2020 this is amazing!

I added a couple of commits for nits (standardise 'diam' to 'diameter') and to update the examples notebook but I agree this is all ready to go 🚀 

Thanks again for contributing, it's great to see my code being used and others take an interest and you are welcome to contribute again in future. You write well, but also apply good SE principles, and hopefully we both learnt some new things. 😄

PR #66, opened 2024-02-11
Simplify HcParams interface

### LINKED ISSUE #2: HcParams leads to a complicated interface
It'd be good to hide a lot of the `HCParams` stuff, maybe by making it an optional argument to the functions that use it, and falling back to default values if it isn't provided. I think the ideal system would be something like:

```python
from archeryutils import score_from_handicap, handicap_from_score, rounds

# Default to current AGB methods
myscore = score_from_handicap(rounds.york, 38)
myhc = handicap_from_score(rounds.portsmouth, 575)
```

Alternative handicap systems could be used by setting a context using fancy `with` methods, or as optional args to the standard functions. Or we could just have a global context setter:

```python
archeryutils.set_handicap_system("some_other_method")
```

jatkinson1000:
Closes #2 

As discussed, `HcParams` is made an optional parameter and we fall back on default values if not otherwise specified.
Needs docstring for `HcParams` making clear what the various contents are.

- [x] Source code updated to address issue
- [x] Style and formatting applied
- [x] Tests written to cover changes
- [ ] Docstrings included/updated in code
- [ ] Project documentation updated as necessary

### DIFF
--- a/archeryutils/classifications/agb_indoor_classifications.py
+++ b/archeryutils/classifications/agb_indoor_classifications.py
@@ -268,7 +268,6 @@ def agb_indoor_classification_scores(
     group_data = agb_indoor_classifications[groupname]
 
     hc_scheme = "AGB"
-    hc_params = hc_eq.HcParams()
 
     # Get scores required on this round for each classification
     # Enforce full size face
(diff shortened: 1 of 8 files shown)


jatkinson1000:
Consider making handicap system an optional argument defaulting to "AGB".
Pro - simplifies system, only need to ask for score & round or handicap & round.
Con - depending on user others may wish to be using AA as an alternative.

I suppose if there is no default then people _have_ to specify scheme, so we are not really losing anything by providing one...?

Question is then what to make the default - for our current purposes I would say new AGB, but there could be an argument that the AA system is more widespread...?

jatkinson1000:
Closing this as discussions in #2 have led to me wanting to take a different route to address this and It will be easier to start again rather than refactor/amend what has been done here.

PR #67, opened 2024-02-15
Typing Improvements

### LINKED ISSUE #61: Type hinting improvements
Typing could be made clearer in the project by replacing several Union types for inputs and return values, and replacing them with generic type variables to make it explict that eg floats as input give floats as output, arrays in give arrays out. 

Also the imported container types (Tuple, List, Dict) can be replaced with the builtin types, since these are deprecated since python 3.9

TomHall2020:
Closes #61 

Should be compatible with python 3.9+ syntax. In order of increasing intensity:

- Replaced Dict, Tuple, List etc.
- Use type variable in handicap equations to show when floats or arrays are given as input and are returned as output
- type hints for fixed dictionaries for AGB classification data that are read from JSON
- enforce correct scoring systems for target class using literal types.

The sum of the last three points allow for removing some of the extra assertions that were in place just to please mypy, especially in the score_for_round functions.

There are still a couple of Anys floating around that I might be able to excise, but this feels like most of it, and enough that I could do with a check over to see if anything else needs included in this one.

On the topic of making the scoring systems explicitly typed, I did think about looking at handicap systems as well, either as a literal type or an Enum, but with potential refactors coming from #2 that seemed like a step too far.

Other general feedback from the process of working on this:
- Some of the JSON files being read in are extremely small and I question the value of having to read them in every time they are used vs just defining them in code, in particular the AGB Genders (which the functions to build the classification dictionaries assume the contents of anyway).
- Type hints in the test files were a bit of a pain in the backside and a few needed to be changed/ignored, which adds a bit more friction to development.

### DIFF
--- a/archeryutils/classifications/agb_field_classifications.py
+++ b/archeryutils/classifications/agb_field_classifications.py
@@ -13,7 +13,6 @@ agb_field_classification_scores
 
 import re
 from typing import TypedDict
-import numpy as np
 
 from archeryutils import load_rounds
 import archeryutils.classifications.classification_utils as cls_funcs
@@ -259,11 +258,6 @@ def agb_field_classification_scores(
     # Get scores required on this round for each classification
     class_scores = group_data["class_scores"]
 
-    # Make sure that hc.eq.score_for_round did not return array to satisfy mypy
-    if any(isinstance(x, np.ndarray) for x in class_scores):
-        raise TypeError(
-            "score_for_round is attempting to return an array when float expected."
-        )
     # Score threshold should be int (score_for_round called with round=True)
     # Enforce this for better code and to satisfy mypy
     int_class_scores = [int(x) for x in class_scores]
--- a/archeryutils/classifications/agb_field_classifications.py
+++ b/archeryutils/classifications/agb_field_classifications.py
@@ -63,134 +63,43 @@ def _make_agb_field_classification_dict() -> dict[str, GroupData]:
         "3rd Class",
     ]
 
+    agb_field_scores = {
+        ("Compound", "Male", "Adult") : [393, 377, 344, 312, 279, 247],
+        ("Compound", "Female", "Adult") : [376, 361, 330, 299, 268, 237],
+        ("Recurve", "Male", "Adult") : [338, 317, 288, 260, 231, 203],
+        ("Recurve", "Female", "Adult") : [322, 302, 275, 247, 220, 193],
+        ("Barebow", "Male", "Adult"): [328, 307, 279, 252, 224, 197],
+        ("Barebow", "Female", "Adult"): [303, 284, 258, 233, 207, 182],
+        ("Longbow", "Male", "Adult"): [201, 188, 171, 155, 137, 121],
+        ("Longbow", "Female", "Adult"): [303, 284, 258, 233, 207, 182],
+        ("Traditional", "Male", "Adult"): [262, 245, 223, 202, 178, 157],
+        ("Traditional", "Female", "Adult"): [197, 184, 167, 152, 134, 118],
+        ("Flatbow", "Male", "Adult"): [262, 245, 223, 202, 178, 157],
+        ("Flatbow", "Female", "Adult"): [197, 184, 167, 152, 134, 118],
+        ("Compound", "Male", "Under 18") : [385, 369, 337, 306, 273, 242],
+        ("Compound", "Female", "Under 18") : [357, 343, 314, 284, 255, 225],
+        ("Recurve", "Male", "Under 18") : [311, 292, 265, 239, 213, 187],
+        ("Recurve", "Female", "Under 18") : [280, 263, 239, 215, 191, 168],
+        ("Barebow", "Male", "Under 18") : [298, 279, 254, 229, 204, 179],
+        ("Barebow", "Female", "Under 18") : [251, 236, 214, 193, 172, 151],
+        ("Longbow", "Male", "Under 18") : [161, 150, 137, 124, 109, 96],
+        ("Longbow", "Female", "Under 18") : [122, 114, 103, 94, 83, 73],
+        ("Traditional", "Male", "Under 18") : [210, 196, 178, 161, 143, 126],
+        ("Traditional", "Female", "Under 18") : [158, 147, 134, 121, 107, 95],
+        ("Flatbow", "Male", "Under 18") : [210, 196, 178, 161, 143, 126],
+        ("Flatbow", "Female", "Under 18") : [158, 147, 134, 121, 107, 95],
+    }
+
     # Generate dict of classifications
-    # for both bowstyles, for both genders
-    compound_male_adult: GroupData = {
-        "classes": agb_field_classes,
-        "class_scores": [393, 377, 344, 312, 279, 247],
-    }
-    compound_female_adult: GroupData = {
-        "classes": agb_field_classes,
-        "class_scores": [376, 361, 330, 299, 268, 237],
-    }
-    recurve_male_adult: GroupData = {
-        "classes": agb_field_classes,
-        "class_scores": [338, 317, 288, 260, 231, 203],
-    }
-    recurve_female_adult: GroupData = {
-        "classes": agb_field_classes,
-        "class_scores": [322, 302, 275, 247, 220, 193],
-    }
-    barebow_male_adult: GroupData = {
-        "classes": agb_field_classes,
-        "class_scores": [328, 307, 279, 252, 224, 197],
-    }
-    barebow_female_adult: GroupData = {
-        "classes": agb_field_classes,
-        "class_scores": [303, 284, 258, 233, 207, 182],
-    }
-    longbow_male_adult: GroupData = {
-        "classes": agb_field_classes,
-        "class_scores": [201, 188, 171, 155, 137, 121],
-    }
-    longbow_female_adult: GroupData = {
-        "classes": agb_field_classes,
-        "class_scores": [303, 284, 258, 233, 207, 182],
-    }
-    traditional_male_adult: GroupData = {
-        "classes": agb_field_classes,
-        "class_scores": [262, 245, 223, 202, 178, 157],
-    }
-    traditional_female_adult: GroupData = {
-        "classes": agb_field_classes,
-        "class_scores": [197, 184, 167, 152, 134, 118],
-    }
-    flatbow_male_adult: GroupData = {
-        "classes": agb_field_classes,
-        "class_scores": [262, 245, 223, 202, 178, 157],
-    }
-    flatbow_female_adult: GroupData = {
-        "classes": agb_field_classes,
-        "class_scores": [197, 184, 167, 152, 134, 118],
-    }
+    classification_dict = {}
 
-    # Juniors
-    compound_male_u18: GroupData = {
-        "classes": agb_field_classes,
-        "class_scores": [385, 369, 337, 306, 273, 242],
-    }
-    compound_female_u18: GroupData = {
-        "classes": agb_field_classes,
-        "class_scores": [357, 343, 314, 284, 255, 225],
-    }
-    recurve_male_u18: GroupData = {
-        "classes": agb_field_classes,
-        "class_scores": [311, 292, 265, 239, 213, 187],
-    }
-    recurve_female_u18: GroupData = {
-        "classes": agb_field_classes,
-        "class_scores": [280, 263, 239, 215, 191, 168],
-    }
-    barebow_male_u18: GroupData = {
-        "classes": agb_field_classes,
-        "class_scores": [298, 279, 254, 229, 204, 179],
-    }
-    barebow_female_u18: GroupData = {
-        "classes": agb_field_classes,
-        "class_scores": [251, 236, 214, 193, 172, 151],
-    }
-    longbow_male_u18: GroupData = {
-        "classes": agb_field_classes,
-        "class_scores": [161, 150, 137, 124, 109, 96],
-    }
-    longbow_female_u18: GroupData = {
-        "classes": agb_field_classes,
-        "class_scores": [122, 114, 103, 94, 83, 73],
-    }
-    traditional_male_u18: GroupData = {
-        "classes": agb_field_classes,
-        "class_scores": [210, 196, 178, 161, 143, 126],
-    }
-    traditional_female_u18: GroupData = {
-        "classes": agb_field_classes,
-        "class_scores": [158, 147, 134, 121, 107, 95],
-    }
-    flatbow_male_u18: GroupData = {
-        "classes": agb_field_classes,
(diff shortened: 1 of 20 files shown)


TomHall2020:
Apologies, didn't realise my main branch was behind, merge conflicts fixed on my end but worth a double check.

archeryutils/classifications/agb_field_classifications.py
```diff
@@ -332,11 +258,6 @@ def agb_field_classification_scores(
...
     class_scores = group_data["class_scores"]
 
-    # Make sure that hc.eq.score_for_round did not return array to satisfy mypy
-    if any(isinstance(x, np.ndarray) for x in class_scores):
-        raise TypeError(
-            "score_for_round is attempting to return an array when float expected."
-        )
     # Score threshold should be int (score_for_round called with round=True)
     # Enforce this for better code and to satisfy mypy
     int_class_scores = [int(x) for x in class_scores]
```
jatkinson1000, line 263: I _think_ these lines are now superfluous here, given `score_for_round` isn't called and `GroupData` now carries the information that the scores returned are `int`s?

archeryutils/classifications/agb_indoor_classifications.py
```diff
@@ -92,17 +98,24 @@ def _make_agb_indoor_classification_dict() -> Dict[str, Dict[str, Any]]:
...
                         + delta_hc_age_gender
                         + (i - 1) * bowstyle["classStep_in"]
                     )
 
+                groupdata: GroupData = {
+                    "classes": agb_classes_in,
+                    "classes_long": agb_classes_in_long,
+                    "class_HC": class_hc,
+                }
+                classification_dict[groupname] = groupdata
```
jatkinson1000, line 117: Could this be made one statement to remove temporary variable `groupdata`, thereby resolving the `too-many-locals` issue?

archeryutils/classifications/agb_old_indoor_classifications.py
```diff
@@ -247,11 +259,6 @@ def agb_old_indoor_classification_scores(
...
     ]
 
-    # Make sure that hc.eq.score_for_round did not return array to satisfy mypy
-    if any(isinstance(x, np.ndarray) for x in class_scores):
-        raise TypeError(
-            "score_for_round is attempting to return an array when float expected."
-        )
     # Score threshold should be int (score_for_round called with round=True)
     # Enforce this for better code and to satisfy mypy
     int_class_scores = [int(x) for x in class_scores]
```
jatkinson1000, line 264: As above, I _think_ this is now unnecessary.

archeryutils/classifications/agb_outdoor_classifications.py
```diff
@@ -54,6 +65,9 @@ def _make_agb_outdoor_classification_dict() -> Dict[str, Dict[str, Any]]:
     ArcheryGB 2023 Rules of Shooting
     ArcheryGB Shooting Administrative Procedures - SAP7 (2023)
     """
+    # Five too many locals, but better than repeated dictionary assignment => disable
+    # pylint: disable=too-many-locals
+
```
jatkinson1000, line 70: Yes, I think this is much more readable.
Previous code made sense at the time as we were developing the system, but a fresh/alternative pair of eyes is good.
The difficulty of 'scientific' software vs. clearly-defined output software...

jatkinson1000 (commented):
Thanks @TomHall2020 this is great!
A couple of queries/comments and lines I think may now be redundant.

Apologies for not telling you about the docs - I had it on my mind to drop you a message saying you'd want to rebase, but forgot. -_- PS - I wrote some [docs](https://archeryutils.readthedocs.io/en/latest/)!

If you could look at the comments, amend and push with formatting and linting that'd be great. On a side note, how would you feel about using [pre-commit](https://pre-commit.com/) on the repo. In general I'm not that keen on it, but I can see it could be useful on this project?

General queries:

Changes from `Union[int, float]` to `float` are because python is 'clever enough' to take an integer input and just convert it to a float, so as long as we always use floats internally mypy doesn't complain? I think I'm still thinking in Fortran/C after a week of work.

I wonder if the `ScoringSystem` type should live in utils or target.
I think I err on it being in target as you have done given it is directly associated with targets.

jatkinson1000:
Oh, and python 3.9 doesn't have `TypeAlias`, but I am erring on dropping and making the project 3.10+.
This resolves the issue over in #59 and given the current users I don't think is a real issue.
I'd want to do this in a separate PR that you can rebase this on rather than here, though.

jatkinson1000:
Regarding the JSON, yes, it had crossed my mind that reading every time isn't ideal, but it's not a bottleneck.
Some things could be defined in the code, but these files (specifically the `class_func.read_X_json()` functions) are also used to set up a database for archerycalculator and having a single point of reference is nice.
I can see that the information _might_ be moved to constants.py in a different PR but I like to have it defined clearly and concisely in one place rather than being embedded throughout functions/files.
I'll have a ponder.

archeryutils/classifications/agb_field_classifications.py
TomHall2020: Yes, that does sound right. I think I hesitated here because the outdoor/new indoor ones are generated via score_for_round. Which incidentally I tried to type accounting for the round=True with overloads but it defeated me!

archeryutils/classifications/agb_indoor_classifications.py
TomHall2020: I initially tried that but I think to include the type hint for the TypedDict it has to be assigning to a variable, you can't type hint the assignment to a dictionary value.
Possibly it could be done with a cast but that doesn't feel any better.

archeryutils/classifications/agb_outdoor_classifications.py
```diff
@@ -391,8 +407,8 @@ def calculate_agb_outdoor_classification(
     groupname = cls_funcs.get_groupname(bowstyle, gender, age_group)
     group_data = agb_outdoor_classifications[groupname]
 
-    # We iterate over class_data keys, so convert use OrderedDict
-    class_data: OrderedDict[str, Dict[str, Any]] = OrderedDict([])
+    # dictionary ordering guaranteed in python 3.7+
+    class_data = {}
```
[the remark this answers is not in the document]
TomHall2020: This one was satisfying!

TomHall2020:
> Oh, and python 3.9 doesn't have `TypeAlias`, but I am erring on dropping and making the project 3.10+. This resolves the issue over in #59 and given the current users I don't think is a real issue. I'd want to do this in a separate PR that you can rebase this on rather than here, though.

Good spot. It's not necessary, just more explicit. We could install typing_extensions and get access to it, as well as other goodies like `Self` (3.11), though theres not that much else I've found myself looking for. Otherwise can just remove and add back in after a bump to 3.10.

archeryutils/targets.py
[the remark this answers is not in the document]
TomHall2020: This is a matter of taste. Options include: 
1. leaving it as `str` type, 
2. giving it a more explicit alias `ScoringSystem: TypeAlias = str`
3. assigning the literal types as in here
4. creating an Enum of scoring systems and passing that instead of string values
    1. Optionally a StrEnum to allow the values to actually still be strings rather than Enum types, though thats a 3.11 feature

The advantage of options 3 or 4 is that 
1. The supported systems are listed outside of the Target class, so can be introspected easily by someone in an interactive session/notebook/IDE rather than reading the `Target.__init__` source code
2. Type checking picks up invalid scoring systems right at the point you initialise a Target or Pass. Especially handy for still using strings (I've tried to make a `worcster` target so many times).

TomHall2020:
> On a side note, how would you feel about using [pre-commit](https://pre-commit.com/) on the repo. In general I'm not that keen on it, but I can see it could be useful on this project?

I've never had to use it before but aware of it; I can't really offer any experience for an opinion, but I guess the question is the benifit of keeping every commit clean worth forcing all future contributors be comfortable/able to use it before they can add anything to the project? 

> Changes from `Union[int, float]` to `float` are because python is 'clever enough' to take an integer input and just convert it to a float, so as long as we always use floats internally mypy doesn't complain? I think I'm still thinking in Fortran/C after a week of work.

Less that it converts it to a float but just that ints are compatible with operations on floats. [mypy docs](https://mypy.readthedocs.io/en/latest/duck_type_compatibility.html) explain it, essentially so long as you don't explicitly call any float only methods you are generally safe to assume ints are usable anywhere floats are. 

> I wonder if the `ScoringSystem` type should live in utils or target. I think I err on it being in target as you have done given it is directly associated with targets.

Agree here, makes sense to keep it close to where its used, and importable from the same place as Target (any consumer wanting to write a type checked function that creates Targets would need the ScoringSystem type too).

TomHall2020:
> Regarding the JSON, yes, it had crossed my mind that reading every time isn't ideal, but it's not a bottleneck. Some things could be defined in the code, but these files (specifically the `class_func.read_X_json()` functions) are also used to set up a database for archerycalculator and having a single point of reference is nice. I can see that the information _might_ be moved to constants.py in a different PR but I like to have it defined clearly and concisely in one place rather than being embedded throughout functions/files. I'll have a ponder.

Makes sense, I didn't consider those files being read in by other projects.

Out of scope for this but perhaps a good compromise is to just call the read_json functions once in a module (either classification_utils or a seperate classifcation_constants), assign the results to constants and then import those in each of the modules they are used. 

Actually has the side benifit of moving those to global variables rather than locals in the _make_classification_dict functions.

TomHall2020:
Thats annoying, I did run black to generate the final commit there, but it didn't fix the spacing of module docstrings to imports on my local version

```zsh
❯ black archeryutils
reformatted /Users/Tom/dev/forks/archeryutils/archeryutils/classifications/agb_field_classifications.py
reformatted /Users/Tom/dev/forks/archeryutils/archeryutils/classifications/agb_outdoor_classifications.py
reformatted /Users/Tom/dev/forks/archeryutils/archeryutils/rounds.py
reformatted /Users/Tom/dev/forks/archeryutils/archeryutils/targets.py
reformatted /Users/Tom/dev/forks/archeryutils/archeryutils/handicaps/handicap_equations.py
reformatted /Users/Tom/dev/forks/archeryutils/archeryutils/tests/test_rounds.py
reformatted /Users/Tom/dev/forks/archeryutils/archeryutils/tests/test_targets.py
reformatted /Users/Tom/dev/forks/archeryutils/archeryutils/handicaps/tests/test_handicaps.py

All done! ✨ 🍰 ✨
8 files reformatted, 21 files left unchanged.

❯ which black
/Users/Tom/dev/forks/archeryutils/.venv/bin/black

❯ black --version
black, 23.12.1 (compiled: yes)
Python (CPython) 3.11.5
```

And now I see that the version of black in the project requirements got bumped and I hadn't reinstalled the dependencies, so that explains it.

jatkinson1000:
> Out of scope for this but perhaps a good compromise is to just call the read_json functions once in a module (either classification_utils or a separate classification_constants), assign the results to constants and then import those in each of the modules they are used.
> 
> Actually has the side benefit of moving those to global variables rather than locals in the _make_classification_dict functions.

Please open an issue (no obligation to take it on, but I agree this would be better so should document it.)
It would be an internal change I think, so not required for release.

archeryutils/targets.py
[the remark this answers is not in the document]
jatkinson1000: Yep, happy with that.
Looks like `TypeAlias` [was deprecated in python 3.12](https://peps.python.org/pep-0695/), so unless there is a particular reason to use `TypeAlias = Literal[...]` as opposed to `Literal[...]` I think the latter is fine, and has the benefit that it satisfies tests in 3.9 (so the comment about bumping to 3.10 can be removed).

jatkinson1000 (commented):
OK, I think this is pretty much good to go.

One thing would be if you could take a look over the docs and examples notebook to see if you feel anything needs updating as a result of these changes.
(It may well be that the answer is no as much of this is internals).
Possible things from a quick glance are:
- Target docstring - does it need the type updating to Literal/ScoringSystem for `system`?
- Rounds can now take iterable of passes, could have an example in the docstring for this, at least the tuple?

I added some general guidance on PRs [here](https://archeryutils.readthedocs.io/en/latest/develop/contributing.html#opening-a-pull-request) that includes this, but this work predates that, so don't worry about format for this.

My last query is about the apparent deprecation of `TypeAlias` in 3.12 (see comment) and whether we want to leave the code as is here (and remove the comment about bumping in v3.10+)?

archeryutils/targets.py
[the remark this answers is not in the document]
TomHall2020: The depreciation of `TypeAlias` is because its replaced with the new typing syntax:
```python
MyType = str  # python 3.9
MyType: TypeAlias = str  # 3.10-3.11
type MyType = str  # 3.12+
```
Its helpful to be explicit that these things are type aliases, otherwise it can be unclear if something is doing variable assignment or something else. Which is why the second version exists, the 3.12 type syntax is even cleaner but we're a long way from that so I think its still helpful to leave the comment given a 3.10 bump is on the immediate horizon? And then the comment can be removed and the future change for 3.12 can just be done by grepping for TypeAlias

TomHall2020:
Figuting out how to get sphinx to play nicely with displaying the allowed values is defeating me here, so I'm not going to get too drawn into it. I've found from playing that if I ~~put an empty line between the brackets~~ don't provide the type in the docstring sphinx is autofilling the allowed literal values when it builds the docs:

````
    scoring_system : 
        target face/scoring system type. Must be one of the supported values.
```
Gives this:
<img width="748" alt="Screenshot 2024-02-16 at 17 22 44" src="https://github.com/jatkinson1000/archeryutils/assets/72264908/4e1d5e35-1a76-4d74-9de5-45a9f87a4da0">

But clearly this is not much use in the actual docstring.

I've left it as something that isn't quite as nicely formatted in the html docs, but fits with the suggested numpy docstring syntax, doesn't autofill and duplicate its values, and shows the relevant information in the docstring itself.
```
    scoring_system : {\
        "5_zone", "10_zone", "10_zone_compound", "10_zone_6_ring",\
        "10_zone_5_ring", "10_zone_5_ring_compound", "WA_field", "IFAA_field",\
        "IFAA_field_expert", "Beiter_hit_miss", "Worcester", "Worcester_2_ring"}
        target face/scoring system type. Must be one of the supported values.
```

archeryutils/targets.py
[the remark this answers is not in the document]
jatkinson1000: Ahhh, that makes sense, I did feel it would be unusual to just remove it outright rather than upgrade.
Yes, I think I will merge this as is on this topic, but will bump to 3.10 soon.
I've certainly learnt a few things about (up to date) typing from this!

jatkinson1000 (approved):
Looking good to go, thanks so much.

Added a couple of nits, docstring for Pass, and made ScoringSystem docstring look nicer with rest systax.

Will merge shortly (just considering best merge format) 🚀

Issue #15, opened 2023-03-13
Round object has to be defined with a name


jatkinson1000:
At present a `Round` object has to defined with a name and a list of Passes.
However, for abstract use we may just want to define passes and no name.
Consider a refactor to make `name` an optional argument.

Will have knock on effects where `Round` is used as calls will need to change.
Breaking change?

Requires:
- [ ] Review how much of a breaking change this will be
- [ ] Update Rounds.py `Round` class
- [ ] Update Rounds tests
- [ ] Update knock on effects in rest of code
- [ ] Update any archerycalculator code affected
- [ ] Notify of changes in docs?

jatkinson1000:
After working to implement this I think this is actually not an issue and should be left as it is.

The resulting code for pre-defined rounds is less intuitive with a mass of passes required before setting a name:
```
york = Round(
    [
        Pass.at_target(72, "5_zone", 122, (100, "yard"), False),
        Pass.at_target(48, "5_zone", 122, (80, "yard"), False),
        Pass.at_target(24, "5_zone", 122, (60, "yard"), False),
    ],
    "York",
)
hereford = Round(
    [
        Pass.at_target(72, "5_zone", 122, (80, "yard"), False),
        Pass.at_target(48, "5_zone", 122, (60, "yard"), False),
        Pass.at_target(24, "5_zone", 122, (50, "yard"), False),
    ],
    "Hereford",
)
```

I'm also not convinced that there is much case to define without a name - if using we will always be assigning the `Round` to a variable, so why not assign a `"name"` at least the same as the variable name.*
And in, e.g. an app, users creating custom rounds to be added to a database would want to give them a name so they can re-use them in future.

\* Only case I can see is if defining the round as part of the the function call to elsewhere, but I would argue this is bad usage of the code.

I will let this sit for a little longer but will close on consideration after the current WIP MRs are complete.

jatkinson1000:
Closing as discussed above.

PR #72, opened 2024-02-18
Class improvements - constructors and reprs

### LINKED ISSUE #70: Feature Request: Define Pass using existing Target
Often we define a target before defining a Pass, especially in the examples.
However, we then pass exactly the same information to the `Pass` constructor that internally re-makes a `Target`.
It would be nice, if we have already constructed a `Target`, to pass this directly to an alternative `Pass` constructor.

May need to be careful with ensuring that we use a copy rather than a reference in the `Pass`...

```python
import archeryutils as au

mytarget = au.Target("10_zone", (122, "cm"), (70.0, "m"))

mypass = au.Pass(36, mytarget)
```

TomHall2020:
Fairly small patch this one but closes #70 and improves the QOL of working with the classes a little.
- Add reprs to Target, Pass and Round
- Change Round constructor to directly accept a Target instance.
   - Add an `at_target` classmethod to replicate the previous functionality of passing parameters through to create a target
   - Implementation is exactly the same as previous init function
- Updated documentation and examples notebook

Optional change as its slightly out of scope, had to touch the load_rounds module to adapt to new constructors, while I was there cleaned up the read_json function a little by:
- factoring out the various string aliases for locations
- swapping manual dict iteration for missing diameter units for a get call with default value "cm"

### DIFF
--- a/archeryutils/rounds.py
+++ b/archeryutils/rounds.py
@@ -19,24 +19,15 @@ class Pass:
     ----------
     n_arrows : int
         number of arrows in this pass.
-    scoring_system : {\
-        ``"5_zone"`` ``"10_zone"`` ``"10_zone_compound"`` ``"10_zone_6_ring"``\
-        ``"10_zone_5_ring"`` ``"10_zone_5_ring_compound"`` ``"WA_field"`` ``"IFAA_field"``\
-        ``"IFAA_field_expert"`` ``"Beiter_hit_miss"`` ``"Worcester"`` ``"Worcester_2_ring"``}
-        target face/scoring system type. Must be one of the supported values.
-    diameter : float or tuple of float, str
-        face diameter in [centimetres].
-    distance : float or tuple of float, str
-        linear distance from archer to target in [metres].
-    indoor : bool, default=False
-        is round indoors for arrow diameter purposes?
+    target : Target
+        A Target object.
 
     Attributes
     ----------
     n_arrows : int
         number of arrows in this pass.
     target : Target
-        A Target object defined using input parameters.
+        A Target object.
 
     Examples
     --------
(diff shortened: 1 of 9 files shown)


docs/getting-started/quickstart.rst
```diff
@@ -78,9 +78,15 @@ a target:
 
 .. ipython:: python
 
-    my70mPass = au.Pass(36, "10_zone", 122, 70.0)
+    my70mPass = au.Pass(36, my720target)
     print(my70mPass.max_score())
 
+We can also directly construct a Pass and Target at the same time using the `at_target` constructor
```
jatkinson1000, line 84: I know it's pedantic, but I'd prefer a slight rephrase here as we technically don't construct a `Target` as a separate entity, rather we bypass the need to create an intermediate `Target` to provide to `Pass` by providing the target data directly to create a `Pass`.

examples.ipynb
```diff
@@ -174,7 +174,25 @@
...
-    "my70mPass = au.Pass(36, \"10_zone\", 122, 70.0)"
+    "my70mPass = au.Pass(36, my720target)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "15415889",
+   "metadata": {},
+   "source": [
+    "We can also directly construct a Pass and Target at the same time using the `at_target` constructor"
```
jatkinson1000, line 185: Phrasing as per comment in docs.

TomHall2020:
Finally, hit the checks first time. 

The constructors are a little cleaner this way round than my first go in #59 (I had to mess around with `object.__new__` there to sneak the Target in).

If you fancy while this is open we could also add support for equality comparisons. Right now it defaults to comparison by identity, its a bit jarring that you can make two Target objects with exactly the same parameters and they won't compare equal. Plus they can still be used as dict keys (defaulting to hash by identy), which is even more confusing:

```python
>>> my_target = Target("5_zone", 122, 50)
>>> my_lookup_data = {my_target : "Something useful"}

# somewhere else later on...
>>> reassigned_target = Target("5_zone", 122, 50)
>>> my_lookup_data[reassigned_target]
KeyError: Target('5_zone', (1.22, 'metre'), (50, 'metre'), indoor=False)
>>> # wtf?
>>> my_lookup_data
{Target('5_zone', (1.22, 'metre'), (50, 'metre'), indoor=False): 'Something useful'}
>>> # uh-oh
>>> my_target == resassigned_target
False
```

Wheras if we implement __eq__ and not __hash__ then they will become unhashable by default which at least reduces the potential for nonsense.

Can open a seperate issue to discuss if you want, I think this would be the easiest place to include the change to fix it though. Or just merge this one in and come back to it later, I don't mind.

archeryutils/targets.py
```diff
@@ -127,6 +127,21 @@ def __init__(
         self.native_dist_unit = Length.definitive_unit(native_dist_unit)
         self.indoor = indoor
 
+    def __repr__(self) -> str:
```
jatkinson1000, line 130: I wonder if it would be better to write out in native units?\
A user usually calls `repr()` to check and get an understanding, in which case they'd probably be thinking in the native units they defined the target in, rather than in the underlying working units?

Happy to hear thoughts otherwise.

archeryutils/rounds.py
```diff
@@ -19,24 +19,15 @@ class Pass:
...
 
     Attributes
     ----------
     n_arrows : int
         number of arrows in this pass.
     target : Target
-        A Target object defined using input parameters.
+        A Target object.
 
     Examples
```
jatkinson1000, line 32: The examples here need updating to use the new `__init__` method (rather than replacing please move them to `at_target` which currently lacks examples).

ll. 184 (example for `Round`) can also be updated to use the new notation (worth a check over to see if I missed any others anywhere...).

jatkinson1000 (changes requested):
Yep, pretty much happy with this.

A nit on the docs phrasing, and couple of missed docstring updates, and a query about target `repr`.

Happy to add equality support here if you want to - I see no reason not to and it makes sense to go in with __repr__.

archeryutils/targets.py
TomHall2020: Actually this does make much more sense. I wanted to avoid sprinkling more unit conversion logic around, but instead I factored it into some properties and was able to re-use it more in `Round.get_info` too.

TomHall2020:
Should be good to go from my end

jatkinson1000 (approved):
Yep, this is fantastic.

Added a couple more test cases and parameterised the target equality to match pass and round tests.

One thing that came up as I added another test was "should two rounds be considered equal if they have the same passes but in a different order?"\
I think we're getting too deep into rule technicalities here (I believe for AGB the answer is no, but for WA you can do (a FITA at least in) reverse distance order or even more bizarre combinations (e.g. Grand FITA)).
I think for our purposes it's probably worth saying "No - pass order matters" to keep things simple and worry about this later.\
I know we had similar discussions when developing with Rob Potts about when two rounds should be considered the same and the key thing for us was e.g. WA1440-70m and Metric I.
My suggestion would be to merge this as is now, but remove the `name` comparison to capture the 1440-Metric cases as part of #15.

We also discussed if the ruleset matters - this was at a time when the timing rules for WA had changed but AGB had not, such that there now was a distinction between the WA1440-70m and Metric I. I note that the equality methods on rounds implemented here do not account for `location`, `body`, or `family`, so would not catch this. I think this is OK, and could be changed in future as an internal I think.

Let me know your thoughts, and I'll then go ahead and merge.

TomHall2020:
Yeah, I had this thought as well, I'd considered starting to look at sorting the passes and then seeing if they were the same or reversed copies of each other, but I think its all getting too specific. 
Really the purpose of the equality comparison at API level I think is to check that you successfully recreated an object that will quack the same. I'd suggest if we want to support detailed governing body specific rules then that's a case for business logic in higher level functions, that could implent mappings of round aliases for predefined rounds, and maybe reach into the round classes, order their passes and check them directly etc.

So in short I think we should keep the name sensitive comparison here and worry about it elsewhere.

Issue #49, opened 2024-01-08
Upload to pypi


TomHall2020:
Is there much else that needs to be done before this can be uploaded? 

Am currently working on a project using this and being able to properly install and version it rather than installing from source would be quite helpful. 

Happy to try and contribute if anything needs doing to prepare for this.

jatkinson1000:
I could probably look at doing this in its current state with a bit of tidying to be honest (done is better than perfect etc.etc.)
I was hoping to improve the field handicaps, but have been too busy to do much recently so this may become something for a later version (and it's all unofficial at the moment anyway).

Always keen to see how it's useful.
How is your project packaged at the moment, as there are ways to add git repos to dependencies and point to specific commits/hashes? This could work in the interim. See [here](https://github.com/jatkinson1000/archerycalculator/blob/main/pyproject.toml) for an example.

jatkinson1000:
@TomHall2020 Having refreshed my memory I think I'd like to close #2, #4, and #27 for release, as implementing these in future would likely be breaking changes, so best to set now.

#3 and #15 would be nice, but I think can be resolved in a backwards compatible way in future.

If you fancy taking a crack at #27 please do! - I have set out a rough plan for what would need doing in the issue.
Happy to help with guidance on forks/pull requests etc. if required.
I'll try and look at #4 and #2 this weekend depending on how much I freeze at a frostbite/EFAA course.

TomHall2020:
Yeah, #27 looks pretty straightforward I can cook something up for that. 
The project isn't packaged as such, its a django web app so archeryutils is just a dependancy on the server side. At the moment the person I'm working on it with has just included this project as a submodule but the git pointer in requirements.txt would probably work better for us for now.

TomHall2020:
Hate to add to the list but theres probably one more issue to go before release, #74

jatkinson1000:
Yeah, it's been a bugbear of mine for a while, I'll do it after #2 as it should hopefully be relatively straightforward with some default arguments.

jatkinson1000:
Revisiting this, and once we close #57 (PR ready), #71 (PR #82 WIP), and #56 (PR #59 close) I think we are good to go.

jatkinson1000:
#57 is now closed, #56 is now closed, #71 is only one remaining.

PR for PyPI release has been opened in #87

PR #78, opened 2024-02-23
Handicaps refactor

### LINKED ISSUE #2: HcParams leads to a complicated interface

### LINKED ISSUE #60: Make some handicap_functions methods private
All except `handicap_from_score` and `﻿﻿print_handicap_table` should be private.
To do this requires removing the explicit tests for any private functions and instead testing them implicitly through the above two.

This will require tests writing for `print_handicap_table` which should bump coverage up.

jatkinson1000:
OK, this has been a slog and I will need a bit of a break after, but here it is.\
In the words of Dylan Beattie: https://www.youtube.com/watch?v=xCGu5Z_vaps

MR Checklist:
- [X] Tests written to cover changes
- [X] Docstrings included/updated in code
- [x] Project documentation updated as necessary

Summary:
- The handicaps code has been fully refactored into a class structure, with each scheme being a subclass of a base `HandicapScheme` class
- Handicap table functionalities have also been brought inside a `HandicapTable` Class
- A set of wrapper functions around the `HandicapScheme` classes have been written to provide easy access and preserve similar behaviour of the previous API.

This will close #2 
This inadvertently also closes #60 

TODO:
- [x] Update examples notebook
- [x] Add a function to `HandicapTable` to yield a string output for use elsewhere (archerycalculator).
- [x] Decide how best to add the specific documentation for each `HandicapScheme` subclass to the documentation.


jatkinson1000:
@TomHall2020 and @LiamPattinson please do take a look and let me know your comments.
If you would be happy to review please let me know and I'll add you.

TomHall2020:
Yeah would be happy to review on this. Will pull down the fork and play with it interactively too, but the new public API does look way better (good riddance `HcParams`!)

archeryutils/handicaps/handicap_scheme_aa.py
```diff
@@ -0,0 +1,296 @@
...
+        self.params = {
+            "ang_0": ang_0,  # Baseline angle used for group size 1.0 [millirad].
+            "k0": k0,  # Offset required to set handicap 100 at desired score.
+            "ks": ks,  # Change with each step of geometric progression.
+            "kd": kd,  # Distance scaling factor [1/metres].
+        }
+
+        self.arw_d_out = 5.0e-3
+
+        self.name = "AA"
```
LiamPattinson, line 96: I don't really like how `name`, `arw_d_out`, and `desc_scale` are class variables in the super class but instance variances in the subclass. It means `HandicapAA.name == "unnamed"`, but `HandicapAA().name == "AA"`. I'd recommend either keeping these as class variables or as instance variables throughout. To have them as instance variables, you could set them in the super class `__init__` function, then start every subclass `__init__` with `super().__init__()`.

Just overriding the class variables in the subclasses would work too, or if you wanted to get fancy and make them (pseudo-)immutable you could implement them as properties. 

You could also get really weird and set the class variables using [`__init_subclass__`](https://www.geeksforgeeks.org/__init_subclass__-in-python/) and set them in subclasses like:

 ```python
class HandicapAA(HandicapScheme, name="AA", arw_d_out=5.0e-3, desc_scale=False):
    ...
```
I think it'd look something like the following in the base class:

```python

def __init_subclass__(cls, name: str, arw_d_out: float=5.5e-3, arw_d_in: float=9.3e-3, desc_scale: bool=True):
    super().__init_subclass__() # not sure if this is needed?
    cls.name = name # must be set in subclasses
    cls.arw_d_in = arw_d_in # optional
    cls.arw_d_out = arw_d_out # optional
    cls.desc_scale = desc_scale # optional
```

It might be overkill, but the advantage is similar to that of `@abstractmethod` in that it forces subclasses to set a name. The main disadvantage is that it's a fairly obscure bit of Python and might just cause developer confusion further down the road.

archeryutils/handicaps/handicap_scheme_aa.py
```diff
@@ -0,0 +1,296 @@
...
+
+    def __init__(
+        self,
+        ang_0: float = 1.0e-3,
+        k0: float = 2.37,
+        ks: float = 0.027,
+        kd: float = 0.004,
+    ):
+        self.params = {
+            "ang_0": ang_0,  # Baseline angle used for group size 1.0 [millirad].
```
LiamPattinson, line 88: I can't see why the parameters are stored within the dict `self.params` rather than just being regular variables. Is it just to ensure all subclasses have the same contents?

archeryutils/handicaps/handicap_scheme.py
```diff
@@ -0,0 +1,764 @@
...
+    sigma_r(handicap, dist)
+        Calculate radial deviation for a given handicap and distance.
+    arrow_score(target, handicap, arw_d=None)
+        Calculate the average arrow score for a given target and handicap.
+    score_for_passes(rnd, handicap, arw_d=None, rounded_score=True)
+        Calculate the expected score for all passes in a round for a given handicap.
+    score_for_round(rnd, handicap, arw_d=None, rounded_score=True)
+        Calculate the expected score for a round for a given handicap.
+    handicap_from_score(score,rnd, arw_d=None, int_prec=False)
+        Calculate the handicap for a given score on a given round.
```
TomHall2020, line 74: I think it might make more sense to settle on round as the first positional argument to `handicap_from_score`, given `score_for_passes` and `score_for_round` both take it first.

archeryutils/handicaps/handicap_scheme.py
```diff
@@ -0,0 +1,764 @@
...
+        else:
+            handicap = 175.0
+            delta_hc = -0.01
+
+        # Set rounding limit
+        if self.name in ("AGB"):
+            round_lim = 1.0
+        else:
+            round_lim = 0.5
+
```
TomHall2020, line 591: It feels odd that the method on the base class is doing conditional logic looking for a specifc handicap systems name here. 
I'd suggest refactor so that the rounding limit is passed as a parameter to this function or stored as another system variable to be looked up?

archeryutils/handicaps/handicap_scheme.py
```diff
@@ -0,0 +1,764 @@
...
+        """
+        max_score = rnd.max_score()
+
+        if self.desc_scale:
+            handicap = -75.0
+            delta_hc = 0.01
+        else:
+            handicap = 175.0
+            delta_hc = -0.01
```
TomHall2020, line 584: Dispatching on desc scale is coupling the base method to the specifics of the AGB and AA systems. As with the rouding limit I think its better to make these look up named variables on the schemes themselves, perhaps something like:
```
# In AGB subclasses
absolute_minimum_handicap=-75.0
desc_scale = True

# in AA subclasses
absolute_minimum_handicap=175.0
desc_scale = False

# in this method
handicap = self.absolute_minimum_handicap
delta_hc = 0.01 * (-1 if not self.desc_scale else 1)
```

Obviously naming etc to taste but I think it makes sense to decouple.

archeryutils/handicaps/handicap_scheme_aa.py
TomHall2020: Yeah, I also don't yet understand why the indirection here. A quick search on the project and I think `.params` is only accessed in the class specific `sigma_t` methods, which means that every method can reasonably know about what parameters it should be able to access.

archeryutils/handicaps/handicap_functions.py
```diff
@@ -12,754 +12,314 @@
...
-    # Check we have a valid score
-    max_score = rnd.max_score()
-    if score > max_score:
-        raise ValueError(
-            f"The score of {score} provided is greater than the maximum of {max_score} "
-            f"for a {rnd.name}."
-        )
-    if score <= 0.0:
+    if isinstance(handicap_sys, HandicapScheme):
+        return handicap_sys
```
TomHall2020, line 90: I really like this interface, returning the same scheme if you already happen to be holding on to one, but taking the strings as well. Very neat.

archeryutils/handicaps/handicap_functions.py
```diff
@@ -12,754 +12,314 @@
...
-    ...     999,
-    ...     wa_outdoor.wa1440_90,
-    ...     "AGB", hc_params,
-    ...     int_prec=True
-    ... )
-    44.0
+    >>> from archeryutils import handicaps as hc
+    >>> agb_scheme = hc.handicap_scheme("AGB")
+    >>> agb_scheme
+    <archeryutils.handicaps.handicap_equations.HandicapAGB object at 0x31415ee00>
```
TomHall2020, line 86: This looks a bit ugly showing up in the docstring, I think it shows that the Schemes need a nicer repr.
Something like `<HandicapScheme("AGB")>` `<HandicapScheme name="AGB">`, formatted to taste?

archeryutils/handicaps/handicap_scheme_aa.py
jatkinson1000: Nice spot, thanks! - I have been working on this late at nights so missed that!
They should indeed be set as instance variables rather than class variables (I think is my preference), so I'll put an `__init__` in `HandicapScheme` and then call the super from the subclass `__init__`.

archeryutils/handicaps/handicap_scheme.py
```diff
@@ -0,0 +1,764 @@
...
+            return self._get_max_score_handicap(rnd, arw_d, int_prec)
+
+        handicap = self._rootfind_score_handicap(score, rnd, arw_d=arw_d)
+
+        # Force integer precision if required.
+        if int_prec:
+            if self.desc_scale:
+                handicap = np.ceil(handicap)
+            else:
+                handicap = np.floor(handicap)
```
TomHall2020, line 526: Same point on coupling of `desc_scale` to implementation specific AA/AGB rules. I think desc_scale should only be used to influence searching upwards or downards on the handicap scale in these algorithms, but should definitly avoid keying other behaviour off it.

archeryutils/handicaps/handicap_scheme.py
```diff
@@ -0,0 +1,764 @@
...
+        # too many statements (64/50), branches (17/12), and variables(23/15).
+        # However, it is a single enclosed algorithm => disable
+        # pylint: disable=too-many-locals
+        # pylint: disable=too-many-branches
+        # pylint: disable=too-many-statements
+
+        if self.desc_scale:
+            x_init = [-75.0, 300.0]
+        else:
+            x_init = [-250.0, 175.0]
```
TomHall2020, line 652: Same point about coupling desc_scale to specifics of handicap system search parameters

archeryutils/handicaps/handicap_scheme.py
jatkinson1000: Agreed, I'd made a mental note to look at this one, but it wasn't as straightforward as the other cases where I swapped out the checks using names for the `desc_scale` attribute.

TomHall2020:
Had to stop my review there, don't know if it was the diffs or my browser but the github web interface just started going nuts. Most of what I've picked up on is the implicit coupling of `desc_scale` to handicap system specific data in the base HandicapScheme class, should be fairly easy to fix. Otherwise absolutely way hapier with the API, I can build a round and get useful information on it in a REPL so much more easily than before

```python
from archeryutils import Round, Pass, Target
from archeryutils import handicaps as hc

t = Target("5_zone", (48, "inch"), (80, "yards"))
p = Pass(30,t)
r = Round('Test', [p,p,p])

hc.score_for_round(r, 30, "AGB")
hc.handicap_from_score(700, r, "AGB")
```

Much more intuitive than the previous version and I appreciate the reduced imports as well.

archeryutils/handicaps/handicap_scheme_aa.py
LiamPattinson: I can understand it if the user wants to know what their current setup is, as calling `my_scheme.params` is much easier than parsing the results of `vars(my_scheme)` and trying to figure out which variables relate to the scheme. I'm not sure how likely this is to happen though, and removing the indirection would make the code in `sigma_t` a little cleaner. It's really not a huge deal either way though.

archeryutils/handicaps/tests/test_handicaps.py
```diff
@@ -80,65 +74,63 @@
...
+            match=(
+                "InvalidSystem is not a recognised handicap system.\n"
+                + "Please select from 'AGB', 'AGBold', 'AA', 'AA2'."
+            ),
+        ):
+            hc.handicap_scheme("InvalidSystem")
+
+    @pytest.mark.parametrize(
+        "scheme",
+        # Check all systems, different distances, negative and large handicaps.
```
TomHall2020, line 101: Leftover comment doesn't apply to this test

archeryutils/handicaps/handicap_scheme_aa.py
TomHall2020: Could always generate that dict of params as a property if you want it easily exported. It just feels like an unusual pattern to me to pass them as individual instance variables but then access them bundled through the dict. If I was looking to to contribute to the project with a new handicap scheme that would certainly have be scratching my head a little.

archeryutils/handicaps/handicap_scheme_aa.py
jatkinson1000: https://github.com/jatkinson1000/archeryutils/pull/78/commits/8e5440f7af3a87edd8bb4a49901bd483bedca504 should fix this I think.

archeryutils/handicaps/handicap_scheme_aa.py
jatkinson1000: Passing in as separate variables means you could change/override just one value without having to also pass all of the others in, and if you did want to pass multiple you could do it with a `kwargs` dict.

e.g. I can see an instance where people want to amend the 'zero handicap' position without changing the general scaling - this was indeed something I played around with whilst developing the scheme trying to get a scratch point that was at 1400 on a 1440 or 700 on a 720.

I don't think this one needs overthinking as 99.9% of users are unlikely to ever touch things at this depth in the code. Having it as a dict may come in useful when writing a `__repr__` method...

archeryutils/handicaps/handicap_scheme.py
jatkinson1000: Yes, one think I need to do as part of this MR is standardise variable names and order - I noticed as I consolidated that there were discrepancies between `round_score_up` and `rounded_score` etc. that both performed the same function.

archeryutils/handicaps/tests/test_handicaps.py
jatkinson1000: https://github.com/jatkinson1000/archeryutils/pull/78/commits/fac281c9ceca4541960a00ce009b6d84f50846a6 should resolve this.

archeryutils/handicaps/handicap_scheme.py
jatkinson1000: I'm inclined to make the standardised order `handicap`, `round`, as this would match the `sigma_r/t()` which uses `handicap`, `distance`. Not sure it's massively important as long as it gets standardised in some way.

jatkinson1000: This _should_ be dealt with in https://github.com/jatkinson1000/archeryutils/pull/78/commits/851a98cc2d12233e1087da98216007033844d37e but please do cast an eye as I could easily have missed something.

jatkinson1000: Actually, it would probably make sense to also standardise the classification functions to also take `score` then `round`? And I need to update the docs etc.

archeryutils/handicaps/handicap_functions.py
jatkinson1000: Should be dealt with in https://github.com/jatkinson1000/archeryutils/pull/78/commits/42c67412e1b0c60a71b222b9deea8e73a3ebd704
Toyed with displaying more info, but user can always get that with more probing, so decided to adhere to KISS and use `HandicapScheme('AGB')`.

jatkinson1000: OK, hopefully resolved with https://github.com/jatkinson1000/archeryutils/pull/78/commits/116947e5c439e8135f1cda4af17a5733d42d0a59 to address comments on the above commits.
Also added `__repr__` and `__str__` for `HandicapTable` in https://github.com/jatkinson1000/archeryutils/pull/78/commits/05b23c2cacb97be5af0ec44302e61884c33b9b9b

archeryutils/handicaps/handicap_scheme.py
jatkinson1000: I think this is OK, as the decision on whether you round to the integer above or below is dependent on whether the handicap scale is ascending or descending.

archeryutils/handicaps/handicap_scheme.py
jatkinson1000: Do you feel https://github.com/jatkinson1000/archeryutils/pull/78/commits/6ce0f78fd8a4ee47e0946cb96d199f30a25be49e resolves this?
I think it can still be set on `desc_scale`, as the direction of the search will determine whether we are searching from low to high or vice versa, but I have removed the implicit dependence on specific schemes to make the start point set from a handicap scheme attribute (`scale_bounds`).

jatkinson1000: I also think the implemented approach is slightly clearer as it indicated whether we are searching from the max or min bound in each case, rather than having a variable `min_handicap` that takes a maximal numerical value.

TomHall2020: Yeah, that does look better. If I'm being picky I'd still prefer to see as little data set as possible in the base class, rather just the type declarations on the variables and forcing all subclasses to set their data explicitly (makes it **much** easier to track down that data too if its all in one place), so I think it would be better for `desc_scale` and `scale_bounds` (and the `max_score_rounding_lim`) to be set explicity in HandicapAGB and HandicapAGBold. Even with the helpful comment above the variable declarations in the base, this relys a bit too much on implicit inheritance when the data could be explicitly defined where its applied.

archeryutils/handicaps/handicap_scheme.py
TomHall2020: Yeah, on re-reading I twigged that this time, any scheme with an ascending/descending scale would need to choose ceil or floor appropraitely

archeryutils/handicaps/handicap_scheme.py
jatkinson1000: To be set as an instance variable in the base class (required to satisfy the code that uses them in the base class methods) they have to be declared and given a value (I don't think we can have uninitialised variables in python).

I do get what you're saying however, so options might be:
- Leave as is, but explicitly overwrite every variable in each classification subclass to reinforce.
- Initialise with junk (`0.0`, 'True' etc.) and explicitly overwrite every variable in each classification subclass.
  but I'm not keen on placing a junk default there in case.
- Outfit the code using [pydantic](https://docs.pydantic.dev/latest/) which would allow implementing as described (plus perhaps some [other bonus features](https://www.youtube.com/watch?v=C-GCX9Ftzko)).

archeryutils/handicaps/handicap_scheme_aa.py
jatkinson1000: What do you think about #79 which would neaten this up?

archeryutils/handicaps/handicap_scheme.py
jatkinson1000: Had a go at the pydantic approach - What do you think to #79?

TomHall2020: Ah, thats annoying, my comment here never posted as it got lost in a pending review!!

I was trying to say theres an easier way, you dont *need* pydantic for this:

```python
class HandicapScheme(ABC):
    desc_scale: bool    
    scale_bounds: list[float]
    max_score_rounding_lim: float
```

Just declare the variables with type annotations (dataclass style); mypy is happy, no junk variables, can then overwrite in every subclass

jatkinson1000: 😅 Having been through it I do like the neatness of the pydantic approach in #79 which cleans up some things such as not having an `__init__` and neat addition of other scheme-specific parameters.
The extra checks it brings are also nice from a security point of view, but could be argued as overkill given much of this is internal rather than user-facing code.

TomHall2020: I do agree the code looks better after the diff there, but at the cost of a bit of technical debt. I feel the need to figure out what magic things pydantic is or isn't doing if I want to change anything in that part of the code, and I'm not personally familiar with it. I think we can get it nearly as clean without it so personally I wouldn't bother with the extra dependency. But also its pretty small scale on a part of the code I'm unlikely to touch so I wouldn't straight up oppose it.

jatkinson1000: > ```python
> class HandicapScheme(ABC):
>     desc_scale: bool    
>     scale_bounds: list[float]
>     max_score_rounding_lim: float
> ```
> 
> Just declare the variables with type annotations (dataclass style); mypy is happy, no junk variables, can then overwrite in every subclass

In this case they're being declared as class attributes rather than instance attributes, however, which is what @LiamPattinson raised as an issue in https://github.com/jatkinson1000/archeryutils/pull/78#discussion_r1500868639. Whilst we could then overwrite as class attributes in the subclasses this perhaps smells a bit off/is a bit obscure - a value that has to be set specifically for every subclass.
After sleeping on it pydantic may also help with type validation for `Target` and `Round`, allowing failure at initialisation rather than in a deeper routine at runtime...

LiamPattinson: I don't use MyPy very often, and I tend to use type annotations just as in-code documentation, so I'm not totally clued up on this stuff, but I think that the following doesn't declare class variables:

```python
class HandicapScheme(ABC):
    desc_scale: bool    
    scale_bounds: list[float]
    max_score_rounding_lim: float
```

Since you don't assign to them, no class variables are created, and only static type checkers pay attention to those lines. They'll be interpreted as instance variables that will be set at some point, and to instead get an annotation for a class variable you need to write `myclassvar: ClassVar[int]`.

jatkinson1000: Huh, OK.
That said, I'm not super keen on this as:
- It doesn't readily seem super clear whether they are class or instance variables.
- It doesn't really create items which are used as part of the class methods.

TomHall2020: This was in my original comment that got swallowed by the review, I didn't rewrite it but you can also totally declare the variable types in the base class init if that is more to your tastes:

```
class HandicapScheme(ABC):
    def __init__(self, *args):
        self.desc_scale: bool    
        self.scale_bounds: list[float]
        self.max_score_rounding_lim: float
```

I have a diff that literally just changes this, declares types in the base init, declares all relavent variables in the subclass inits (and removes the redundant type annotations there), passes all tests and satisfies mypy too.
```
git diff --stat
 archeryutils/handicaps/handicap_scheme.py     | 10 +++++-----
 archeryutils/handicaps/handicap_scheme_aa.py  |  6 ++++--
 archeryutils/handicaps/handicap_scheme_agb.py | 14 +++++++++++++-
 3 files changed, 22 insertions(+), 8 deletions(-)
```

archeryutils/handicaps/handicap_scheme_aa.py
jatkinson1000: @LiamPattinson Is this resolved to your satisfaction?

archeryutils/handicaps/tests/test_handicaps.py
jatkinson1000: @TomHall2020 Is this resolved to your satisfaction?

archeryutils/handicaps/handicap_scheme.py
jatkinson1000: @TomHall2020 Is this now resolved to your satisfaction?

archeryutils/handicaps/handicap_scheme.py
jatkinson1000: OK, this should be implemented in https://github.com/jatkinson1000/archeryutils/pull/78/commits/d7ac0ec01a1906ec077b95576f6e29eb1782488d Are we all now happy with this implementation?

archeryutils/handicaps/handicap_scheme.py
jatkinson1000: Standardising classification equations to take score as first argument implemented in https://github.com/jatkinson1000/archeryutils/pull/78/commits/f30fc02d6ccf008e8f8b792695b25846b180d909

Are you happy with how things are now set up/standardised @TomHall2020?

jatkinson1000:
@LiamPattinson @TomHall2020 I have updated the docs which was the last TODO from my end.
I believe I have also addressed the points raised above in review.

If I can get closure of those and an approving review I will merge this which should unblock a number of your PRs @TomHall2020 and bring us very close to completing #49.

archeryutils/handicaps/handicap_scheme.py
TomHall2020: Looking at the last patch all good from me

archeryutils/handicaps/handicap_scheme.py
TomHall2020: Consistent is good, played with it interactively and can settle/get used to it so all happy for me

TomHall2020 (approved):
Looks good to me, fundamentally all the issues I had are addressed, the public api is much improved, and should be a clean foundation to work from in the future.

archeryutils/handicaps/handicap_scheme_aa.py
jatkinson1000: As detailed over in #79 that has been abandoned and sticking to form as discussed above.

jatkinson1000:
OK, great.

If you are happy please could you do a review with "Approve" selected @TomHall2020?
(Go to Files Changed, select green review button, add a message and then select the "Approve" radio button instead of "Comment".)
That should then unlock it for me to merge (I put branch protection rules on now that there is more than just me opening PRs).

archeryutils/handicaps/handicap_scheme_aa.py
jatkinson1000: This is resolved, and in-fact somewhat moot as all instance variables are now explicitly set in each subclass so I will mark as resolved.

TomHall2020:
> If you are happy please could you do a review with "Approve" selected @TomHall2020? (Go to Files Changed, select green review button, add a message and then select the "Approve" radio button instead of "Comment".) That should then unlock it for me to merge (I put branch protection rules on now that there is more than just me opening PRs).

I've reviewed and approved already, but from the status bar below it looks like the repo is set to require an approving review from someone with write access, so that might be the issue you're having.

LiamPattinson (approved):

LiamPattinson:
I've added my approval and I'm seeing the same issue: at least 1 approving review is required by reviewers with write access.

jatkinson1000:
Thanks folks - feels great to get this one finally checked off!

@TomHall2020 if you can update your open PRs to use this (I know you already rebased one) then hopefully they can follow soon.

PR #79, opened 2024-02-24
Use pydantic for HandicapScheme

jatkinson1000:
This MR to the ongoing #78 implements pydantic as a way of tackling some of the issues raised in review.
Specifically that of taking numerical parameters out of the base class making it truly generic and implementing the 

Pro:
- Makes the code more secure and intuitive
- May make setting of scheme specific parameters even easier

Con:
- Adds a dependency, but useful.
- Makes it harder for people to contribute as they may need to get up to speed on pydantic

Ambivalent:
- There is probably much more that could be done with a move to be a Pydantic model.
  Should we leverage some of that?
  Is pydantic overkill for what we are trying to achieve here?

TODO:
- [ ] Check how we now set/override certain parameters in the child classes.
- [ ] Update the `handicap_scheme()` function to work with this.
- [ ] Update docs and docstrings

### DIFF
--- a/archeryutils/handicaps/handicap_scheme.py
+++ b/archeryutils/handicaps/handicap_scheme.py
@@ -33,17 +33,18 @@ References
 
 import warnings
 from abc import ABC, abstractmethod
-from typing import Optional, TypeVar, Union, overload
+from typing import Optional, Sequence, TypeVar, Union, overload
 
 import numpy as np
 import numpy.typing as npt
+from pydantic import BaseModel
 
 from archeryutils import rounds, targets
 
 FloatArray = TypeVar("FloatArray", float, npt.NDArray[np.float_])
 
 
-class HandicapScheme(ABC):
+class HandicapScheme(BaseModel, ABC):
     """
     Abstract Base Class to represent a generic handicap scheme.
 
@@ -57,11 +58,13 @@ class HandicapScheme(ABC):
         diameter of an indoor arrow [metres] for this scheme, default 9.3e-3
     desc_scale: bool
         does the scheme work on a descending scale i.e. lower handicap is better, default True
-    scale_bounds: list[int]
+    scale_bounds: Sequence[int]
         Reasonable upper and lower bounds on the handicap scale for bounding searches
     max_score_rounding_lim: float
         Limit to round the max score to when searching
         depends on scheme rounding method e.g. round() vs. ceil() etc.
+    scale_bounds: list[int]
+        Reasonable upper and lower bounds on the handicap scale for bounding searches
 
     Methods
     -------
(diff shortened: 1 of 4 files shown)


jatkinson1000:
@TomHall2020 How about this as a solution to your comments in #78?

Opening as a subbranch so it can be merged in this MR if we like it, but can be discarded easily if not.
Discussion points noted in the description.

@LiamPattinson feel free to also cast an eye if you feel like it.

LiamPattinson:
I tend to lean against Pydantic, as it adds a lot of mysterious behind-the-scenes machinery that I don't fully understand (something about metaclasses I think?), and it can impose significant performance penalties. If I ever need to enforce type/value constraints at runtime, I normally just use custom properties or [descriptors](https://docs.python.org/3/howto/descriptor.html).

On the other hand, I know a lot of people swear by it, and I've heard that the later releases have improved the performance a lot. I quite like the constrained types it adds too, like `conint` and `confloat`, which lets you say, "This is a float in the range [0, 1]", or, "This int is strictly greater than zero".

No strong feelings either way on this one.

TomHall2020:
I'm becoming more strongly -1 to this the more I look at it.

I can see the percieved advantages of some slightly cleaner looking code, but I've yet to have any practical experience of implementing it sucessfully in a small project, I think it will add as much friction as it helps.

Concrete issues I have are:
1. Clutter from subclassing on the objects
```python
from archeryutils import handicaps as hc
AA = hc.handicap_scheme('AA')

# extra listings on dir(AA) when it is a pydantic base model
{'__annotations__',
 '__class_getitem__',
 '__class_vars__',
 '__copy__',
 '__deepcopy__',
 '__fields__',
 '__fields_set__',
 '__get_pydantic_core_schema__',
 '__get_pydantic_json_schema__',
 '__getattr__',
 '__iter__',
 '__pretty__',
 '__private_attributes__',
 '__pydantic_complete__',
 '__pydantic_core_schema__',
 '__pydantic_custom_init__',
 '__pydantic_decorators__',
 '__pydantic_extra__',
 '__pydantic_fields_set__',
 '__pydantic_generic_metadata__',
 '__pydantic_init_subclass__',
 '__pydantic_parent_namespace__',
 '__pydantic_post_init__',
 '__pydantic_private__',
 '__pydantic_root_model__',
 '__pydantic_serializer__',
 '__pydantic_validator__',
 '__repr_args__',
 '__repr_name__',
 '__repr_str__',
 '__rich_repr__',
 '__setstate__',
 '__signature__',
 '_calculate_keys',
 '_check_frozen',
 '_copy_and_set_values',
 '_get_value',
 '_iter',
 'construct',
 'copy',
 'dict',
 'from_orm',
 'json',
 'model_computed_fields',
 'model_config',
 'model_construct',
 'model_copy',
 'model_dump',
 'model_dump_json',
 'model_extra',
 'model_fields',
 'model_fields_set',
 'model_json_schema',
 'model_parametrized_name',
 'model_post_init',
 'model_rebuild',
 'model_validate',
 'model_validate_json',
 'model_validate_strings',
 'parse_file',
 'parse_obj',
 'parse_raw',
 'schema',
 'schema_json',
 'update_forward_refs',
 'validate'}
```

2. Automagic data parsing can cause as many gotchas as it solves. See [gist: unexpected parsing floats to datetimes](https://gist.github.com/mpkocher/30569c53dc3552bc5ad73e09b4852492)

3. Depenency issues. Requiring every project that uses archeryutils to now have a transative dependency on pydantic is objectively, more surface area for issues. We pin to at least v2 here, what if a project is using v1? 

4. Performance. I know its small scale, but it takes about 5 times longer to create a new HandicapScheme object when pydantic is revalidating all the data, that 99.999% of the time is the same data we statically declared in its own constructor.

In something like FastAPI you are buying into their use of pydantic as the backend, and if you don't like it you can choose to use a different framework. Archeryutils is a small utility library but it is also the reference implementation for the current AGB handicap system; there isn't really another option for me to choose from short of forking it.

jatkinson1000:
Slept on it and I am inclined to agree that this is currently overkill for what we want to achieve here.

I am happy with the way it is now implemented over in #78 

Therefore closing this .

On the plus side I've now had practical experience with pydantic which may well come in useful in the future.

PR #80, opened 2024-02-25
Remove warning spam at package import

### LINKED ISSUE #74: UserWarning spam on module import
This one shouldn't be hard but definitely needs fixing, at the moment loading the default rounds raises a ton of UserWarnings for not setting round families. Given this happens at module import time and the user has absolutely no choice about it, this is pretty annoying and fills up the terminal in interactive sessions, test output etc.

Since its fine to instantiate a round directly in the API without specifying family, location etc, this seems like a dev/debug hint, so better to suppress them and have a way to opt back in for dev builds, or just decide if the data is in a good state now and get rid of the warnings in `read_json_to_round_dict` entirely?

jatkinson1000:
Closes #74

There are currently many warnings which appear at import as default rounds are read from file without certain inputs.
This is undesireable and needs cleaning up.

This MR implements this with minimal changes by adding a `verbose` option to the routine that reads rounds in from a json dict. If enabled the warnings are printed.

For consideration:
- The option would be to remove them entirely, but I can see instances where this would be useful.
- This never faces the user though as these are run at import to create dotdicts, so do we need them?
- The defaults when an argument is absent should probably be made to match those of `Round` (i.e. `None`) rather than what is given here (e.g. `"custom"`).

- [ ] Tests written to cover changes
- [ ] ~Docstrings included/updated in code~
- [ ] ~Project documentation updated as necessary~

### DIFF
--- a/archeryutils/load_rounds.py
+++ b/archeryutils/load_rounds.py
@@ -43,7 +43,9 @@ LOCATIONS = {
 }
 
 
-def read_json_to_round_dict(json_filelist: Union[str, list[str]]) -> dict[str, Round]:
+def read_json_to_round_dict(
+    json_filelist: Union[str, list[str]], verbose: bool = False
+) -> dict[str, Round]:
     """
     Read round information from a json file into a dictionary of rounds.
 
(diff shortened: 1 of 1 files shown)


jatkinson1000:
@TomHall2020 How do you feel about this in relation to #80?

TomHall2020:
Yeah, seems like the simplest resolution. The `"body"` and `"family"` checks could be trivially replaced with `body=round_i.get("body")` etc as we do nothing fancy with these, but otherwise makes sense to me. I tried to think of something we could do in testing to flag missing family data in rounds, but coudn't think of anything. 

I guess its really down to code review and maintenance to make sure the data is good when changes/new contributions are made to the supplied rounds, which would have been necessary anyway as no automated test would have the domain knowledge to know what numbers of arrows/distances/round names are appropraite anyway.

jatkinson1000:
Thanks.
Regarding replacing the checks, I implemented this, but decided that I'd prefer to leave as it currently is because this a) matches the rest of the checks in the code and b) is a little clearer what we are doing (`None` is assigned explicitly rather than implicitly).

I am happy for this to go in as is since the overall coverage is improved, and the untested line in the patch was also not tested in the original (testing from json files is more involved (but doable) and "ItS aLL wOrKinG fOR nOw").

If you can drop a review I'll aim to get it merged soon.

PR #82, opened 2024-03-06
Update Python Versions

### LINKED ISSUE #71: Bump to Python 3.10
It has come up in a couple of PRs (#67 #59) that features of python 3.10+ would be useful.
It is a shame to be held back from using features/have to implement messy conditional workarounds.

Whilst 3.9 is still 'supported' by python, the scientific python [SPEC0](https://scientific-python.org/specs/spec-0000/) suggests dropping support towards end of 2023.

Therefore it does not seem unreasonable to raise minimum support to 3.10

jatkinson1000:
Closes #71 

- Raises minimum python version to 3.10 as discussed in #71 
- Also adds testing using latest python 3.12 and uses 3.12 for linting/coverage
- Updates github workflows to latest versions

Checklist:

- [X] ~Style and formatting applied~
- [X] ~Tests written to cover changes~
- [X] ~Docstrings included/updated in code~
- [X] Project documentation updated as necessary

### DIFF
--- a/archeryutils/classifications/agb_indoor_classifications.py
+++ b/archeryutils/classifications/agb_indoor_classifications.py
@@ -292,7 +292,7 @@ def agb_indoor_classification_scores(
     # Handle possibility of gaps in the tables or max scores by checking 1 HC point
     # above current (floored to handle 0.5) and amending accordingly
     for i, (score, handicap) in enumerate(
-        zip(int_class_scores, group_data["class_HC"], strict=False),
+        zip(int_class_scores, group_data["class_HC"], strict=True),
     ):
         next_score = hc.score_for_round(
             np.floor(handicap) + 1,
--- a/archeryutils/classifications/agb_indoor_classifications.py
+++ b/archeryutils/classifications/agb_indoor_classifications.py
@@ -193,7 +193,7 @@ def calculate_agb_indoor_classification(
 
     groupname = cls_funcs.get_groupname(bowstyle, gender, age_group)
     group_data = agb_indoor_classifications[groupname]
-    class_data = dict(zip(group_data["classes"], all_class_scores))
+    class_data = dict(zip(group_data["classes"], all_class_scores, strict=True))
 
     # What is the highest classification this score gets?
     # < 0 handles max scores, > score handles higher classifications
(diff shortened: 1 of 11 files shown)


jatkinson1000:
@TomHall2020 I think this is ready to be merged if you fancy casting an eye?
It updates a number of the 3.10 things that came up during your recent PRs.

archeryutils/classifications/agb_indoor_classifications.py
```diff
@@ -292,7 +292,7 @@ def agb_indoor_classification_scores(
     # Handle possibility of gaps in the tables or max scores by checking 1 HC point
     # above current (floored to handle 0.5) and amending accordingly
     for i, (score, handicap) in enumerate(
-        zip(int_class_scores, group_data["class_HC"]),
+        zip(int_class_scores, group_data["class_HC"], strict=False),
```
TomHall2020, line 295: `strict=False` is the default and the previous behaviour for zip so technically not needed, but if you want to have it here to be extra explicit then no problem.

archeryutils/handicaps/handicap_scheme.py
```diff
@@ -43,18 +43,6 @@
 
 FloatArray = TypeVar("FloatArray", float, npt.NDArray[np.float64])
 
-# itertools.pairwise not available until python 3.10
```
TomHall2020, line 46: good riddance!

archeryutils/handicaps/handicap_scheme.py
```diff
@@ -266,7 +254,7 @@ def _s_bar(
 
         return max_score - sum(
             score_drop * np.exp(-(((arw_rad + (ring_diam / 2)) / sig_r) ** 2))
-            for ring_diam, score_drop in zip(ring_sizes, score_drops)
+            for ring_diam, score_drop in zip(ring_sizes, score_drops, strict=True)
```
TomHall2020, line 257: Doesn't hurt to have the strict check here in case of future fiddling with the implementation, although from how the current version works the two iterators will always be of length N and N + 1 - 1 = N.

TomHall2020 (approved):
Yeah looks good, there's the option to clean up some of the type hinting syntax but that isn't necessary and depends a bit on taste anyway, plus theres always more stuff coming in future versions.

archeryutils/classifications/agb_indoor_classifications.py
jatkinson1000: This should be set to True, though, so will amend now - see https://github.com/jatkinson1000/archeryutils/pull/82/commits/f3c830afee6b30816c0eafb960ebfdff5fa88865

archeryutils/handicaps/handicap_scheme.py
jatkinson1000: Done to satisfy Ruff B905: https://docs.astral.sh/ruff/rules/zip-without-explicit-strict/

PR #84, opened 2024-03-08
Prevent empty Round passes and type guarding

### LINKED ISSUE #57: Bug: Possibility to define a round with no passes
It is possible to define a round with no passes leading to undefined behaviour when subsequently used.
It is unlikely one would do this intentionally, but perhaps there should be a check that the input parameter is a list of `Pass` (or a single `Pass`?).

jatkinson1000:
Closes #57 

- Added a check to raise `ValueError` if empty iterable is passed to `Round` constructor.
- Also added some guards against `archeryutils` types passed to constructors.

Checklist:

- [X] Tests written to cover changes
- [X] ~Docstrings included/updated in code~
- [X] ~Project documentation updated as necessary~

### DIFF
--- a/archeryutils/rounds.py
+++ b/archeryutils/rounds.py
@@ -42,6 +42,9 @@ class Pass:
     def __init__(self, n_arrows: int, target: Target) -> None:
         self.n_arrows = abs(n_arrows)
         self.target = target
+        if not isinstance(self.target, Target):
+            msg = "The target passed to a Pass should be of type Target."
+            raise TypeError(msg)
 
     @classmethod
     def at_target(  # noqa: PLR0913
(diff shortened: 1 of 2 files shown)


archeryutils/rounds.py
```diff
@@ -211,6 +214,12 @@ def __init__(  # noqa: PLR0913
     ) -> None:
         self.name = name
         self.passes = list(passes)
+        if all(False for _ in self.passes):
```
TomHall2020, line 217: We check that the contents are actual passes later so this might be more readable as:
```python
if len(self.passes) == 0:
   ...
```
Which is more explict about what we are checking for, but then that begs the question of why not just do the traditional python approach:
```python
if not self.passes:
```

TomHall2020:
Seems fair, simple fixes.

archeryutils/rounds.py
jatkinson1000: Huh, could've sworn I tried that and it had a problem but seems to work now 🤷 Amended and improved test in https://github.com/jatkinson1000/archeryutils/pull/84/commits/8f3dacfc1c66dce2764359a23a8b8c2b617edcae.

PR #91, opened 2024-05-21
Class fraction

jatkinson1000:
Adds the functionality of giving the 'fraction through a classification' a score is based on the handicap.

As requested by users.


jatkinson1000 (commented):
Should probably use type of classification in the function name for consistency elsewhere. No other schemes can use this at present, but field will need it in #46

archeryutils/classifications/tests/test_agb_outdoor.py
```diff
@@ -565,3 +566,187 @@ def test_calculate_agb_outdoor_classification_invalid_scores(
...
+        frac_unrestricted = class_funcs.agb_outdoor_classification_fraction(
+            score=620,
+            roundname="national",
+            bowstyle="recurve",
+            gender="male",
+            age_group="adult",
+            restrict=False,
+        )
+
+        assert frac_unrestricted == 0.4541258975704667
```
LiamPattinson, line 752: I can't comment on the correctness of the PR so far, but in terms of code quality it looks like you've ticked all the boxes.

The only thing bothering me is that there are a few equality tests with floats. These should probably use `assert np.isclose(x, y)`, or `np.testing.allclose(x, y)` which would give more information if the test fails.

archeryutils/classifications/agb_outdoor_classifications.py
```diff
@@ -7,7 +7,7 @@
 agb_outdoor_classification_scores
 """
 
-from typing import Any, Literal, TypedDict, cast
+from typing import Any, Literal, Optional, TypedDict, cast
```
LiamPattinson, line 10: If you're using Python 3.10+, the preferred way to type hint an optional argument is `Type | None` instead of `Optional[Type]`, but type checkers shouldn't really care either way.

jatkinson1000: See https://github.com/jatkinson1000/archeryutils/pull/91/commits/4625170b3647136ba9a689e3cdc0d5937fb2cf44

Could you open an issue for me to check this more widely in the code perhaps?

jatkinson1000:
Need to add a check for when handicap and score can overlap slightly either way at boundaries after being rounded to integers.

Roughly:

- If score is a boundary value (in `classification_scores`) set to... 0.0? 1.0?... 0.0 makes sense I think as 1 occurs when we are at the highest classification we can be.
- Though on this note check what to do when we hit the top bound - should always be 1.0

archeryutils/classifications/tests/test_agb_outdoor.py
LiamPattinson: Sure, can do. From a quick look, I don't think it's a common issue in the tests. I see one instance where `pytest.approx` has been used, which isn't something I was aware of but it looks like it does pretty much the same thing. Maybe that would be more appropriate here?

archeryutils/classifications/agb_outdoor_classifications.py
LiamPattinson: Will do, and there are some related typing updates that could be included too.

archeryutils/classifications/tests/test_agb_outdoor.py
jatkinson1000: Good point - done with a cheeky amend and force to https://github.com/jatkinson1000/archeryutils/pull/91/commits/29fce0101540bff673e56828cc22adabdf24dd71

jatkinson1000 (commented):
Think this is now OK to go.
Consider bumping to 1.1.0 and adding a release, or instead wait until larger changes such as #46 come in.

PR #97, opened 2024-10-29
Add pytest approx to float comparisons in tests

### LINKED ISSUE #93: Exact comparisons with floating points in tests
There are a few instances in the tests where we assert exact equality between floating point numbers. In some cases this makes sense, such as `assert target.diameter == 1.22`. Other cases might be prone to unexpectedly failing in the future due to floating point round-off errors (I've had this happen in other projects due to NumPy/SciPy updates).

jatkinson1000:
Add pytest approx to float comparisons in tests where not currently used and it makes sense to do so.

Closes #93 

Pytest `approx` or `allclose` is in use in most places, but I have had a check through and added a couple more where it might be appropriate.

There are still some instances where it does not make sense (as @LiamPattinson notes in #93) which have been left.
If any further have been missed we can review them should they raise problems in future, but this can hopefully close #93.

### DIFF
--- a/archeryutils/tests/test_length.py
+++ b/archeryutils/tests/test_length.py
@@ -42,7 +42,7 @@ class TestLengths:
     )
     def test_conversion_to_metres(self, value, unit, result):
         """Test conversion from other units to metres."""
-        assert length.to_metres(value, unit) == result
+        assert length.to_metres(value, unit) == pytest.approx(result)
 
     @pytest.mark.parametrize(
         "value,unit,result",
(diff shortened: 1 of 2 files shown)


jatkinson1000:
@LiamPattinson would you be OK to give this a quick review if you have a moment?
Doing a little spring-cleaning on issues.

LiamPattinson (approved):
All looks good to me! I did a quick scan of the tests for any that you missed, and I think you got everything.

jatkinson1000:
So fast! Thanks @LiamPattinson

PR #103, opened 2024-12-09
Outdoor legacy classifications and generic classification API

### LINKED ISSUE #102: Support Legacy AGB classifications for outdoor rounds?
Hello again!

Currently archeryutils supports calculating handicaps in the new and old AGB systems, and can provide classifications and classification score boundaries for Indoor and Outdoor (and now field) rounds under the new system, and also indoor rounds under the old system.

With the new system being essentially fully integrated now, is it possible to provide classification feature parity for outdoor rounds for the previous system as well? I have a use case for this in that we are trying to keep archr agnostic as to which handicap system it uses, so I would like to still be able to calculate classification score thresholds. 

Of course we could already do this by reverse engineering the previous system and working out which handicaps correlate to which classifications, but I'd rather chuck all that in the bin and use archeryutils as a definitive single source of truth for all such matters.

This would at a minimum mean providing two new functions in the classifications namespace: `calculate_agb_old_outdoor_classification` and `agb_indoor_classification_scores`. It might also be a good opportunity to add a simplified generic api similar to the top level handicap functions where the handicap scheme and discipline (eg indoor, outdoor, field) can be passed as string arguments, leaving us with a pair of conviniences functions `calculate_classifcation` and `get_classification_scores`/`classification_thresholds` or something similar.

Happy to help with implementation as ever.

TomHall2020:
First steps towards implementing the pre-2023 AGB outdoor classifications into archery utils. 
Generally I have started by just trying to do the same style of implementation as the other 5 modules in the classifications package. I went for an explict style of declaring the categories (bowstyle/gender/age) and requirements rather than a looped implict method as in the new style, as that fits better with how its done for the old indoor and field requirements, and also it was easier to construct from scratch and visualise what they are.

I've not dug up a definitieve reference to the old rules of shooting yet but from memory and a bit of pouring over some old classification tables I've worked on the following basis:
- Minimum 12 dozen rounds at max distance for GMB/MB (and JMB? much less familiar with the old requirements for juniors)
- Max distance required for Bowman
- Step down 10m/y for each of 1st, 2nd and 3rd class.
- No minimum arrows for bowman or below?

Checklist
- [x] Made required modules and wired in imports
- [x] Convert newer age categories to their equivalents.
- [x] Add checks for distance and 12 dozen rounds
- [x] Fill in placeholder handicap values
- [ ] Add parametrized tests based on actual known scores

Plan is to get this part at least mostly working, and then get to work on the generic API for classifications, but we can begin the discussions on that while this is being done. There are also a few minor inconsistencies and tweaks to improve the existing classifications code that I'd like to work on while I've got the whole lot in the operating theatre so to speak...

- [x] ~multiple nested for loops over age/bowstyle/gender in new system data generation -> replace with `itertools.product`, will reduce indentation and make easier to read without any changes to the underlying code~ see #110 
- [x] agb_old_field returns `"unclassified"` for scores below lowest threshold, where all other modules return `"UC"` resolved in #105
- [ ] mismatched specific long/short forms for age categories that are accepted `'adult', 'under 18'` but `'50+'`? would make more sense to take `over 50`, but this also points to a deeper issue I'll get to later
- [x] ~exposed functions/objects `old_agb_field_classification_scores` and `old_agb_field_classifications` are inconsistent in naming convention with the rest of the package, suggest changing instead to `agb_old_field_classification_scores` etc. (this really got me for a while where I was trying the same parameters in different function by just editing the function name)~ See #107 
- [x] ~importing classifications module as `class_funcs` in test suite when in source code the very similar name `cls_funcs` is used to import classification_utils - suggest changing this to `cf` similar to importing `archeryutils.handicaps as hc`~ See #121 
- [ ] couple of documentation tweaks where the wrong references are made to indoors/outdoors

More generally beyond this it comes to looking at the generic api, which I ~think~ *hope* will be quite simple to implement along similar lines to how we did in #78, with generic functions along the lines of:
```python
from archeryutils import classifications as cf
cf.calculate_classification(score, roundname, bowstyle, gender, age, hc_sys="AGB", discipline="indoor")
```
That then just dispatches to the appropriate module. Currently each module publicly exposes its `calculate_xx_classification` method  and a `xx_classification_scores` method which is required to power the calculation, though the return type with its fill value of `-9999` isn't super friendly to work with. Both of these are also moved up into the overall classifications package namespace along with the module itself and I would plan to leave them there for backwards compatibility.

Theres also a dictionary of the same name as the module which is essentially the compiled raw data needed by the functions, which we take great pains to make sure can never be recreated by deleting the private function that generated it, something I can't quite work out a reason for. It's definitely bit me before earlier on when I first tried to work on the module and was trying to understand the code. I remember using some of the other private functions in the new outdoor classification module which mutate the underlying data in place and then being unable to regenerate the original data. 

A side effect of that is that the namespace is **really** verbose when trying to get around in an interactive shell or do imports, especailly as tab completion will happily jump you into the modules quite easily where you have to repeat yourself over again to find the function you wanted. Even though they are promoted up to the package namespace I find myself accidentally doing this surprisingly often, and also getting confused by the module and then the data dictionary within that module having the same name.

```python
from archeryutils.classifications.agb_out[tab-completed: door_classificaitons] import calculate_agb_outdoor_classification
```

I propose while adding the generic access function, we could simplify the names of the functions within each module, and just re-alias them to their current form when importing them up to the package. Such that each module would contain a `make_classification_dict/data/requirements` function, a `calculate_classification` and a `classification_scores` (or `score_thresholds` function. This would also make life a bit easier when doing the implementation of the generic functions, as we could just match the correct module with the user provided system and discipline, and then call that modules identically named functions to do the calculations.

At any rate thats a decent brain dump of where my thinking is.


TomHall2020:
> but this also points to a deeper issue I'll get to later

As promised, I think fundamentally the fact that everything going into these classification functions aside from a score is a string, is a serious issue. 

Both in terms of what has to happen internally in the code to process these strings:
```python
#Theres lots of string processing and mangling going on to clean up the arguments
/archeryutils/classifications/agb_old_field_classifications.py:
  178:     if age_group.lower().replace(" ", "") in ("adult", "50+", "under21"):
  179          age_group = "Adult"
  180:     elif re.compile("under(18|16|15|14|12)").match(age_group.lower().replace(" ", "")):
  181          age_group = "Under 18"

  190      if (
  191:         bowstyle.lower().replace(" ", "") in ("compound", "recurve")
  192          and "wa_field_24_red_" not in roundname
  193      ):
  194          return "unclassified"
  195      if (
  196:         bowstyle.lower().replace(" ", "")
  197          in (
  198              "barebow",

...
```

And also for the end user as they are trying to figure out how to use the functions
```python
# you are utterly on your own when you are trying to find the right round name or category names...
In [6]: import archeryutils.classifications as cf

In [7]: cf.agb_outdoor_classification_scores('short warwick', 'recurve', 'womens', 'u18') #wrong round, wrong gender, wrong age class...
---------------------------------------------------------------------------
KeyError                                  Traceback (most recent call last)
Cell In[7], line 1
----> 1 cf.agb_outdoor_classification_scores('short warwick', 'recurve', 'womens', 'u18')

File ./archeryutils/classifications/agb_outdoor_classifications.py:543, in agb_outdoor_classification_scores(roundname, bowstyle, gender, age_group)
    540     bowstyle = "Compound"
    542 groupname = cls_funcs.get_groupname(bowstyle, gender, age_group)
--> 543 group_data = agb_outdoor_classifications[groupname]
    545 # Get scores required on this round for each classification
    546 class_scores = [
    547     hc.score_for_round(
    548         group_data["class_HC"][i],
   (...)
    553     for i in range(len(group_data["classes"]))
    554 ]

KeyError: 'u18_womens_recurve'
```
Finally you end up with the frustating situation that if you have already constructed a round object from scratch, or even loaded one from the pre-assigned rounds. You can't actually use it to work out classifications, even though the classification functions will use a fully instantiated round object to do thier own work.

```python
from archeryutils import load_rounds
rnd = load_rounds.AGB_indoor.portsmouth
cf.agb_indoor_classification_scores(rnd, 'compound', 'male', 'under 14')

# TypeError: Unhasable type: 'Round'
```

It is, I think, a bit too far beyond the scope of what I want to get done in this request. But is something I'm keeping in mind as I work on it.

jatkinson1000:
@TomHall2020 a lot to unpack, so will try and place comments here, but please remind me if I miss something.

As you already raise, this should remain a PR for reviving old outdoor classifications, with other amendments made through other pull requests please.

These could include:

- Use of `itertools.product()` Yep, not come across this before, so useful to learn about. Looks like it should do the job of flattening those loops.
- Standardising old field classifications to return `UC`. I think this came about from handling compatibility in archerycalculator. I have patched it in #104 -> #105 
- Naming of old agb field functions to be consistent with old indoor. Patched in #107
- Moving to use `cf` rather than `class_funcs`. I think I am in agreement with this in principle. Is this only a documentation and notebook change?
- Documentation patches that are spotted.

Other points:

- `under 18` vs. `50+` was to match the WA/ArcheryGB terminology (and this was after a midway change from `master`)... Would need some thought on what is most sensible - there is an argument for sticking with the WA/AGB terminology that users would expect.
- That said, it would be useful to have some way of easily checking what the available options are, as this has caught be before and I agree it is irritating digging back around to remind yourself what to use.
- IIRC the reason for not using a `Round` object was because each classification scheme is restricted to a set of named rounds, and I enforce this by checking that the name appears in the dict of rounds instantiated in each classification module. I suppose one could pass in a `Round` and check that it's name appears in this list and I can perhaps see where this might come in useful, but could this not also be resolved by passing `Round.name` to the classification function?

For the linting issues you are seeing I have submitted a patch in #106 which you can rebase off of once merged.

TomHall2020:
> For the linting issues you are seeing I have submitted a patch in #106 which you can rebase off of once merged.

D'oh, once again caught out by linter versioning. I actually hadn't updated my venv with the changes to main since my last contributions.

jatkinson1000:
@TomHall2020 I have just added the handicap thresholds for the various categories.

I pushed them direct to this branch (couldn't get GitHub to play nice with opening a PR from my repo to your fork) so you'll want to pull to get these locally before any further changes to avoid conflicts.

More generally if you rebase off of the latest main on my repo that should hopefully resolve the conflicts and linting issues.

For tests there is a version of the old classification tables available here: https://southleedsarchers.org.uk/wp-content/uploads/2022/03/AGB-Shooting-Admin-Procedures-2020.pdf

jatkinson1000:
Hey @TomHall2020 is there any update on this?

A user has pointed out (See #142) the discrepancy between the latest version of classifications on GitHub and the PyPI release, so I'm keen to get v2.0.0 released with the API changes so things are consistent.

I'd hoped this might go in there, but if you're busy (I know I am!!) then it may need to wait for v.2.1.0

TomHall2020:
> Hey @TomHall2020 is there any update on this?

Hi @jatkinson1000 , just seen this - not using github at work so I've not been keeping track on here. This has fallen behind the main branch a fair bit now but I had a breif look and I've been able to rebase it and resolve conflicts/get the tests passing locally after rebasing, so it shouldn't be a huge stretch to get the functionality working.

TomHall2020:
This is most of the way to a working implementation that mirrors the existing code - working on this has refreshed me on the project structure after some time away and I'm definitely finding the organisation into all the seperate modules awkward and verbose - think theres a clear path to improve that similar to how we handled the handicaps but that feels like a seperate job now.

Coverage missing for the coaxing logic but otherwise lines are nominally hit - I found it quite hard to actually pick some old values out to use as thresholds for the classifications - internet has mostly come up to date with the new system!

archeryutils/classifications/agb_old_outdoor_classifications.py
```diff
@@ -0,0 +1,458 @@
...
+import numpy as np
+import numpy.typing as npt
+
+import archeryutils.classifications.classification_utils as cls_funcs
+import archeryutils.handicaps as hc
+from archeryutils import load_rounds
+from archeryutils.classifications.AGB_data import AGB_ages, AGB_bowstyles, AGB_genders
+from archeryutils.rounds import Round
+
+ALL_OUTDOOR_ROUNDS = load_rounds.read_json_to_round_dict(
```
jatkinson1000: Strictly speaking we'd need to ditch some newer rounds here (AGB 900, 50m BB 720) but I don't think that's worth worrying about as I can't see why anyone would wand to generate identical legacy tables without access to these rounds.

archeryutils/classifications/agb_old_outdoor_classifications.py
```diff
@@ -0,0 +1,458 @@
...
+        (B.LONGBOW, G.FEMALE, A.AGE_UNDER_18): [61, 64, 68, 73, 77],
+        (B.LONGBOW, G.MALE, A.AGE_UNDER_16): [60, 65, 70, 75, 81],
+        (B.LONGBOW, G.FEMALE, A.AGE_UNDER_16): [66, 70, 74, 78, 83],
+        (B.LONGBOW, G.MALE, A.AGE_UNDER_14): [66, 72, 77, 82, 87],
+        (B.LONGBOW, G.FEMALE, A.AGE_UNDER_14): [72, 77, 82, 87, 95],
+        (B.LONGBOW, G.MALE, A.AGE_UNDER_12): [75, 81, 87, 93, 99],
+        (B.LONGBOW, G.FEMALE, A.AGE_UNDER_12): [78, 83, 88, 94, 99],
+    }
+
+    prerequisites = {
```
jatkinson1000: Feels like there should be a less verbose way to assemble these, but it's not the end of the world to have them like this.

jatkinson1000: Adult and Junior `min_dozen` are always identical so should be a variable.

`min_dists` could be pulled from `[90, 70, 60, 50, 40, 30, 20, 15, 10]` though not sure that's necessarily worth pursuing, but some of the repetition between categories could be made clearer

archeryutils/rounds.py
```diff
@@ -272,6 +272,17 @@ def max_distance(self) -> Quantity:
...
+    #     """
+    #     Return the total number of arrows shot on this round.
+    #
+    #     Returns
+    #     -------
+    #     n_arrows : int
+    #         number of arrows in the round
+    #     """
+    #     return sum(pass_i.n_arrows for pass_i in self.passes)
+
```
jatkinson1000: Dead code.

jatkinson1000 (commented):
Thanks @TomHall2020 agree this is looking to be nearly there.

Left you a couple of comments.
Copying the coaxing tests from the current outdoor implementation should be fine and will get the patch coverage.

Will also need something adding to the regression snapshots once we're happy with the implementation.

My go-to for old SAPs with old tables is [a 2016 pdf I got from the Bronte website](https://bronte-archers.co.uk/wp-content/uploads/2022/05/Proc0203ShootingAdminProceduresOct2016-21395.pdf).
Alas the web is "up to date", despite our advice, just in time for the triennial review.

PR #105, opened 2024-12-09
Return codes for old field classifications

### LINKED ISSUE #104: Old field classifications return words not codes for classifications
As pointed out by @TomHall2020 the old field classification functions return words ("Grand Master Bowman", "1st Class", "Unclassified") instead of codes ("GMB", "1C", "UC") like the other schemes.

This arose handling compatibility with archerycalculator, but it would be best to make it consistent here and instead handle it there.

jatkinson1000:
Fixes #104 

Replaces the return array values with codes instead `['GMB', 'MB', 'B', '1C', '2C', '3C', 'UC']`, and any unclassified by `'UC'`.

Tests also updated to match new return values.


jatkinson1000:
@TomHall2020 how do you feel about this?

archeryutils/classifications/agb_old_field_classifications.py
```diff
@@ -191,7 +191,7 @@ def calculate_old_agb_field_classification(
         bowstyle.lower().replace(" ", "") in ("compound", "recurve")
         and "wa_field_24_red_" not in roundname
     ):
-        return "unclassified"
+        return "UC"
```
TomHall2020, line 194: is it worth making this a global constant to keep DRY?

TomHall2020:
Ooh, nice. Gives me a reference to use for the codes to use for the old outdoor system too, and I definitely like `1C` for first class, for some reason I coudln't think of that when I was trying to shorten the codes.

Wondered about assigning a global constant `UNCLASSIFIED="UC"` for DRY purposes, and could also think about making the list of codes a module level constant as well, could be handy for introspection, but those are seasoning. Rest is simple enough, reccomend merging it in.

jatkinson1000:
Debated on this a little as it doesn't match the new setups, but given the differences (an this being somewhat legacy/separate) I think it makes sense. Done both in https://github.com/jatkinson1000/archeryutils/pull/105/commits/215c852f4cd718548bda7e4fae8cb34ed8338e75

PR #106, opened 2024-12-09
Update formatting to satisfy latest checks.

jatkinson1000:
Looks like ruff and mypy have become a little more strict in the latest versions.

This PR fixes that by updating for:
- [X] ruff
  - [X] Sorting imports in `__init__.py`
  - [X] Iterating over a dict using `.items()`
- [x] mypy
  - [ ] ??


jatkinson1000:
@TomHall2020 This patch should fix the ruff and mypy issues you see on your branch in #103 

I'm still getting to the bottom of why mypy was unhappy and ensuring that I am satisfied with the fix (I have gone off mypy a little, or perhaps I need to dial down the settings).

TomHall2020:
Sure thing, I guess if my stuff so far has prompted a few changes to the rest of the codebase we can merge this in and build from there.

PR #107, opened 2024-12-10
Update old field classification naming

jatkinson1000:
To be consistent with old indoor classification naming.


TomHall2020:
Yup, I definitely would have wanted to make that rename as part of the structure of #103 but good to have it locked in place already and I wouldn't have caught renaming the test functions and extra bits of documentation.

PR #110, opened 2024-12-17
Data looping in classification dict generation functions

TomHall2020:
As discussed in #103, the classification data generation functions can be unindented quite nicely by replacing direct iteration over bowstyle/gender/age divisions with the itertools product of those.

Implementing that was very straightforward, but while I was in there I did smell an opportunity to excise another bit of looping, where the class handicap thresholds and minimum distances themselves are built up by manual iteration over the indecies of the classification labels. That took a bit more effort, but I think is much more explicit this way, and the result is that the loop building up the classification data is now pretty much solely packaging things nicely, and has a minimum of inline logic. 

Inside the agb_outdoor `_assign_min_dist` function I could come up with various ways to build up the required indecies based on the number of mb_categories and adjustments etc, but in the end I felt that just declaring it completely explicitly was the most readable and clear way to do it. The use of `np.take` with `mode="clip"` works really nicely to get rid of the try/except index error handling.

There are still some magic numbers (1 and 2) for offsetting the indecies when using the age/gender steps but since I have absolutely no idea why they were there before I can't do much do make them more explicit here. 

No tests broken, so just up to you @jatkinson1000 if you're happy with the new division of logic?

### DIFF
--- a/archeryutils/classifications/agb_field_classifications.py
+++ b/archeryutils/classifications/agb_field_classifications.py
@@ -124,21 +124,22 @@ def _assign_dists(
     age: cls_funcs.AGBAgeData,
 ) -> tuple[npt.NDArray[np.int64], int]:
     """
-    Assign appropriate minimum distance required for a category and classification.
+    Assign appropriate distance required for a category and classification.
 
-    Appropriate for 2024 ArcheryGB age groups and classifications.
+    Appropriate for 2025 ArcheryGB field age groups and classifications.
 
     Parameters
     ----------
     bowstyle : str,
         string defining bowstyle
     age : dict[str, any],
-        dict containing age group data
+        Typed dict containing age group data
 
     Returns
     -------
-    dists : list[float]
-        [minimum, maximum] distances required for this classification
+    tuple
+        ndarray of minimum distances required for each classification for this bowstyle
+        int of maximum distance that is shot by this bowstyle
 
     References
     ----------
@@ -162,6 +163,8 @@ def _assign_dists(
 
     n_classes: int = 9  # [EMB, GMB, MB, B1, B2, B3, A1, A2, A3]
 
+    # EMB to bowman requires a minimum appropriate distance
+    # Archer tiers can be shot at shorter pegs (min dist reduced by 10m for each tier)
     min_dists = np.zeros(n_classes, dtype=np.int64)
     min_dists[0:6] = min_d
     min_dists[6:9] = np.maximum(min_d - 10 * np.arange(1, 4), 30)
--- a/archeryutils/classifications/agb_field_classifications.py
+++ b/archeryutils/classifications/agb_field_classifications.py
@@ -32,7 +32,7 @@ class GroupData(TypedDict):
     classes_long: list[str]
     class_HC: npt.NDArray[np.float64]
     max_distance: int
-    min_dists: npt.NDArray[np.float64]
+    min_dists: npt.NDArray[np.int64]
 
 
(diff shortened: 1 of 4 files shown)


archeryutils/classifications/agb_field_classifications.py
```diff
@@ -166,8 +156,17 @@ def _assign_dists(
...
+    else:
+        min_d, max_d = age["blue"]
+
+    n_classes: int = 9  # [EMB, GMB, MB, B1, B2, B3, A1, A2, A3]
+
+    min_dists = np.zeros(n_classes, dtype=np.int64)
+    min_dists[0:6] = min_d
+    min_dists[6:9] = np.maximum(min_d - 10 * np.arange(1, 4), 30)
+
+    return min_dists, max_d
```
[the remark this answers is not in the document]
jatkinson1000: Ahhh, I see, it replaces code from the triple-loop above.
Still needs docs updating to reflect this. ;)

archeryutils/classifications/agb_outdoor_classifications.py
```diff
@@ -165,8 +153,8 @@ def _assign_min_dist(
 
     Returns
     -------
-    min_dist : int
-        minimum distance [m] required for this classification
+    min_dists : array of int
+        minimum distance [m] required for this category
```
jatkinson1000, line 157:
```suggestion
        minimum distance [m] required for this category for each classification (EMB -> A3)
```

archeryutils/classifications/agb_outdoor_classifications.py
```diff
@@ -182,35 +170,26 @@ def _assign_min_dist(
...
     # All other categories require max dist for B1 and B2 then step down
-    try:
-        return dists[max_dist_index + (n_class - n_mb) - 1]
-    except IndexError:
-        # Distances stack at the bottom end as we can't go below 15m
-        return dists[-1]
+    else:
+        idxs = np.array([0, 0, 0, 0, 0, 1, 2, 3, 4])
+
+    return np.take(dists, idxs + max_dist_index, mode="clip")
```
jatkinson1000, line 192:
```suggestion
    # Extract the relevant distances for each classification from the dists array and return
    return np.take(dists, idxs + max_dist_index, mode="clip")
```

jatkinson1000 (changes requested):
Thanks @TomHall2020 
Itertools definitely makes it cleaner.

I added a couple of comment suggestions to make it clearer to myself (`np.take()` is new to me) and I think one of the docstrings needs updating in the field routines.
Then should be good to go.

More generally, where min_dist functions returns `int`s, I wonder if these should be `float`s.
They are internal functions and in reality it makes not difference to the functioning of the code, but philosophically if we are comparing them to yard values in some places they should perhaps be `float`s.
Any thoughts?

jatkinson1000:
And to answer your question about the magic numbers @TomHall2020, when developing the outdoor classifications we decided to pin the MB handicap for each bowstyle and then set things relative to that. Hence the -2 outdoors and -1 indoors.

This actually turned out to be a real pain in the ass when it came to developing the field classifications, and I wish we had fixed the EMB point, but it seemed like a good idea at the time. IIRC the suggestion was that MB and below was likely to remain fixed, given the desire for a clear "system of progression" at the lower levels and more substantial data, whereas GMB and EMB may need to be tuned over time. Whether this ever happens we'll see...

Would happily discuss over a coffee/beer/etc.

jatkinson1000 (approved):
This is a useful addition so made the changes and rebased on main to merge.

PR #111, opened 2024-12-27
Round narrows

jatkinson1000:
Adds `n_arrows` attribute to rounds as inexpensive and useful to be exposed at the top level rather than calculated from `passes`.


jatkinson1000:
@TomHall2020 FYI: added this little feature as it was desirable for something else I'm doing.  May be useful to you.

PR #125, opened 2025-03-27
[WIP] archeryutils.django

mjtamlyn:
My intention here is to add Django model and form fields to `archeryutils`, to allow for super simple modelling of archery related information, and in a way where it becomes trivial to get handicap or classification data for them.

Structurally, I've created a submodule `archeryutils.django`. Django is a dependency of this submodule, but wouldn't be required for other things. If preferred, I could develop this as a separate module, but there are some tweaks needed in the core codebase to make it work and this felt more collaborative.

### Proposed feature set

- [x] add `archeryutils.django` as a module
- [x] model and form field for `Round`
  - [ ] make deconstruction for migrations work better?
  - [ ] allow an API like `RoundField(rounds=[WA_outdoor, AGB_outdoor_metric, AGB_outdoor_imperial])`
  - [ ] improve the form field so it uses option groups by round family?
- [ ] model and form field for `au.classifications.AGB_ages` (likely based on [django-enumfield](https://pypi.org/project/django-enumfield/))
- [ ] model and form field for `au.classifications.AGB_bowstyles`
- [ ] model and form field for `au.classifications.AGB_genders`
- [ ] `MultiRoundField` - allows you to select a list of rounds, useful for an entry system, might even be good to pass a family to it?

### Projects which can make use of this include:
[SCAYT results system](https://github.com/mjtamlyn/scayt)
[TamlynScore](https://github.com/mjtamlyn/tamlynscore)
[WCA website](https://github.com/mjtamlyn/wallingfordcastle) (for competition entry system, and then being able to pass entry data directly to TamlynScore)

### Guidance
I've not contributed to this before, so I'm not sure what standards I'm expected to match on documentation (both in the code and formally), type hinting, testing etc.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -89,7 +89,11 @@ ipython_config.py
 #   However, in case of collaboration, if having platform-specific dependencies or dependencies
 #   having no cross-platform support, pipenv may install dependencies that don't work, or not
 #   install all needed dependencies.
-#Pipfile.lock
+#
+#   For package development, Pipenv can be useful for local virtualenv management but is not
+#   needed by others.
+Pipfile
+Pipfile.lock
 
 # PEP 582; used by e.g. github.com/David-OConnor/pyflow
 __pypackages__/
--- a/archeryutils/__init__.py
+++ b/archeryutils/__init__.py
@@ -1,6 +1,6 @@
 """Package providing code for various archery utilities."""
 
-from archeryutils import classifications, handicaps
+from archeryutils import classifications, django, handicaps
 from archeryutils.rounds import Pass, Round
 from archeryutils.targets import Quantity, Target
 from archeryutils.utils import versions
@@ -11,6 +11,7 @@ __all__ = [
     "Round",
     "Target",
     "classifications",
+    "django",
     "handicaps",
     "versions",
 ]
--- a/archeryutils/django/fields.py
+++ b/archeryutils/django/fields.py
@@ -1,0 +1,54 @@
+from django.db import models
+
+from archeryutils.rounds import Round
+
+
+class RoundField(models.CharField):
+    description = "Choose from a set of archery rounds"
+
+    def __init__(self, rounds, **kwargs):
+        self.rounds = rounds
+        self._round_dict = rounds  # TOOD: allow more complex round structures to be passed
+
+        # TODO Hack for deconstruction for now
+        if isinstance(rounds, list):
+            self._round_dict = {r: r for r in rounds}
+
+        kwargs.setdefault("max_length", 64)
+        super().__init__(**kwargs)
+
+    def deconstruct(self):
+        name, path, args, kwargs = super().deconstruct()
+        # TODO - round deconstruction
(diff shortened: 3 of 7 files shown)


jatkinson1000 (commented):
Hi @mjtamlyn it took me a while to find chance to look at this as it's been super busy at work.

It would be good to facilitate use of this in various websites. However, my general feeling is that adding django is a large, specific dependency that does not fit with the core of archeryutils. I wonder if it would be better to have another package `archeryutils-web` or `archeryutils-django` for this?

The addition of a `codename` to the `Round` object is really useful though.

Open to discussion on the best way to realise this.

mjtamlyn:
Sure, that makes sense. I'll rework the changes here to be more minimal and set it up as a separate project, probably call it `archerydjango`, as that's snappy enough to import!

I have a bit of a deadline on for the SCAYT project but I can ship it with installs on specific commits for a while until any changes are polished off and released.

TomHall2020:
Hi Marc, sorry I'd meant to get back to you on this as well. I agree with @jatkinson1000 that it makes sense to have the codename made available on the round objects, but I think it might be better to store the actual data characteristics of the round (number of arrows, distance, target size) on the DB rather than the codename. Having said that, I haven't tried this implementation yet and I haven't grokked what its doing here as some kind of round-multiselect.

In any case, seperate package does probably make sense, but let me know when you set that up and I'd be happy to contribute.

mjtamlyn:
@TomHall2020 It's not just about database storage, it's also for other "machine readable" data formats. For example, an export from, say, the WCA website handling entries could say `round: wa720_70`, which could then be read in by TamlynScore to enter the archers, which can re-export it for consumption by a rankings application and/or a club records application. Putting all the information about the round in each place is rather excessive.

TomHall2020:
I agree it seems excessive, perhaps I'm missing something, but my concern with storing by name is that future name changes can now invalidate stored data. e.g `Mens FITA` -> `WA 1440` or `short junior warwick` -> `Warwick 30` are both real world name changes. I guess you could leave the old codename in place but that will gradually drift over time. 

I guess as soon as we publicly make the codenames available, they become fair game, so for example archeryutils could no longer correct typos or rename for internal organisation purposes, at least without warning and a release process.

jatkinson1000:
@TomHall2020 regarding name changes, I did at one point consider having an 'aliases' category (In fact, I think @mjtamlyn, Rob Potts and I discussed this) to try and accommodate the fact that some rounds can have different names for the same composition (1440 70m vs Metric I etc.)

As it is these two are distinct rounds in archeryutils as they have different "governing bodies", but I can see a case where rounds are renamed we create a new "forwarding" attribute that tells anything requesting a `ladies_fita` codename to instead return a `1440_70`.

mjtamlyn:
Yeah there's definitely something to be said about "legacy" names when things change, and I'm not necessarily 100% happy with every name in the file as it is, but ultimately these are to be "machine readable" names rather than "human readable", so they could be called `bananas` for all I care so long as they give us the right object.

PR #130, opened 2025-04-21
Satisfy mypy with cast of return type to FloatArray for _s_bar in handicap calculations.

jatkinson1000:
Cast type of _s_bar in handicap scheme to explicitly be FloatArray to satisfy mypy. Mypy is raising issues as detecting it could theoretically be int as well as float.

### DIFF
--- a/archeryutils/handicaps/handicap_scheme.py
+++ b/archeryutils/handicaps/handicap_scheme.py
@@ -254,10 +254,13 @@ class HandicapScheme(ABC):
         score_drops = (inner - outer for inner, outer in itr.pairwise(ring_scores))
         max_score = max(ring_scores)
 
-        return max_score - sum(
+        s_bar = max_score - sum(
             score_drop * np.exp(-(((arw_rad + (ring_diam / 2)) / sig_r) ** 2))
             for ring_diam, score_drop in zip(ring_sizes, score_drops, strict=True)
         )
+
+        # Perform a cast to return to satisfy typechecker
+        return cast(FloatArray, s_bar)
 
     def score_for_passes(
         self,


LiamPattinson:
From my reading of the function, should the return type instead just be `float`? I can't see how this could result in an array. Replacing `np.exp` with `math.exp` might also help with the type checking.

jatkinson1000:
If you pass it an array of values in `sig_r` then it should return an array of the same length.

I had issues with it thinking that the return could be an int (because of mas score and score_drop) and mixing `float` and `np.float64` that ground my patience down.

The thought of casting everything to numpy arrays has gone through my mind a few times, but I would like to leave it as being able to run with single floats, as that's how the majority of people will use it I expect.
Perhaps there is a comprimise with making this cast all `sig_r` to `np.array` since it is an internal function.

LiamPattinson:
Ah, sorry, I didn't realise `sum(...)` over a bunch of NumPy arrays would work.

I found that the following works, though will need similar changes elsewhere:

```python
    def _s_bar(
        self, target_specs: targets.FaceSpec, arw_rad: float, sig_r: npt.ArrayLike
    ) -> npt.NDArray[np.float64]:
        sig_r = np.asarray(sig_r)

        target_specs = dict(sorted(target_specs.items()))
        ring_sizes = target_specs.keys()
        ring_scores = list(itr.chain(target_specs.values(), [0]))
        score_drops = (inner - outer for inner, outer in itr.pairwise(ring_scores))
        max_score = max(ring_scores)

        return max_score - np.asarray(
            sum(
                score_drop * np.exp(-(((arw_rad + (ring_diam / 2)) / sig_r) ** 2))
                for ring_diam, score_drop in zip(ring_sizes, score_drops, strict=True)
            )
        )
```

- Use of `ArrayLike` over `FloatArray = float | NDArray[np.float64]` as an input also works when passing in something like a list, and it works with `np.asarray(...)`.
- Always returns `npt.NDArray[np.float64]`. If you pass in a plain `float`, it will result in a 0-dim NumPy array of dtype `np.float64`, but due to duck-typing the users likely won't notice.

If you would prefer to continue returning plain `float` when given a `float` as an input, then I can't immediately see any way around `cast`.

jatkinson1000:
Closing as superseded by #133

jatkinson1000:
@LiamPattinson if you get a moment does #133 look sane to you?

PR #131, opened 2025-04-21
Miscellaneous corrections

jatkinson1000:
A PR to collect various corrections spotted whilst going through the code.

- Update outdated module docstrings in handicap_functions to reflect refactored code.

### DIFF
--- a/archeryutils/handicaps/handicap_functions.py
+++ b/archeryutils/handicaps/handicap_functions.py
@@ -1,14 +1,10 @@
 """
 Code providing various functionalities using the archery handicap equations.
 
-Makes use of the basic handicap equations in handicaps.handicap_equations to do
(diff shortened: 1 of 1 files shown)


jatkinson1000:
Consider updating internal round names for `6_zone` to instead of `small`.
Consider updating name of `WA 70m` to `WA 720 (70m)` etc.

mjtamlyn:
👍

jatkinson1000:
@mjtamlyn I'm tempted to separate name changes and updates here until we settle on a definitive new system...

I'd consider trying to do some backwards compatibility mapping for legacy names.

mjtamlyn:
Yeah once all the names are decided we should just do it all in one go. We can also get rid of the duplication on archery calculator at the same time.

PR #132, opened 2025-04-26
Store codename of the round on the object itself.

mjtamlyn:
This change allows the rounds to "self identify" in a machine readable format. Applications can therefore take CSVs or JSON data and look up the identified round.

This extracts the "non-Django" bit out of #125

### DIFF
--- a/archeryutils/load_rounds.py
+++ b/archeryutils/load_rounds.py
@@ -68,6 +68,8 @@ def read_json_to_round_dict(json_filelist: str | list[str]) -> dict[str, Round]:
             data = json.load(json_round_file)
 
         for round_i in data:
+            codename = round_i["codename"]
+
             # Assign location
             if "location" not in round_i:
                 round_i["location"] = None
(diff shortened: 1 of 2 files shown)


jatkinson1000 (changes requested):
Thanks @mjtamlyn this will be useful.

Code-wise, I think it will be important that the dict item and the codename are identical, so I suggest reading in `round_i["codename"]` once to a temp variable, and then using this to assign both the key in `round_dict` and also the `codename` in the constructor call in `read_json_to_round_dict()`.

Housekeeping-wise, `codename` needs adding to the parameters and attributes list in the docstring for the `Round` class.
~It may also need unit testing, though I can perhaps knock that together quickly by modifying an existing one.~ Edit: it appears I do not directly test other optional attributes, so that one's on me...

The mypy error should be resolved if you rebase on main, and I might globally relax the allowed arguments for ruff, but happy for you to put a `# noqa: PLR0913` on the offending line for now.

mjtamlyn:
Changes as requested and rebased :)

jatkinson1000 (approved):
Thanks @mjtamlyn all seems good to me now.

Hopefully it's useful, and feel free to open PRs/discussions for anything else that may be of benefit in the future.

PR #133, opened 2025-04-26
Update type checking to use numpy NDArray and ArrayLike

jatkinson1000:
Use numpy typing closer to intended use as suggested by @LiamPattinson in #130

Take inputs to functions as `ArrayLike`, casting using `np.asarray()` where appropriate.
Return items as `NDArray[float64]` with single values being duck-typed in the majority of cases.
Removes the `FloatArray` type to just use `NDArray[float64]` explicitly.

Resolves a number of issues seen in latest mypy in a more sensible manner than #130 
Also allows us to simplify some of the "type-checker" appeasement throughout the code e.g. castings and overloads.

### DIFF
--- a/archeryutils/handicaps/handicap_scheme.py
+++ b/archeryutils/handicaps/handicap_scheme.py
@@ -34,7 +34,6 @@ References
 import itertools as itr
 import warnings
 from abc import ABC, abstractmethod
-from typing import cast
 
 import numpy as np
 import numpy.typing as npt
@@ -43,12 +42,12 @@ from archeryutils import rounds, targets
 
 
 def _cast_float_array(var_in: npt.ArrayLike) -> npt.NDArray[np.float64]:
-    """Check we can cast to np.float64 array and do so.
+    """Ensure we can cast to a np.float64 array and do so.
 
     Parameters
     ----------
     var_in : ArrayLike
-        input that is an ArrayLik
+        input that is an ArrayLike
 
     Returns
     -------
@@ -61,10 +60,11 @@ def _cast_float_array(var_in: npt.ArrayLike) -> npt.NDArray[np.float64]:
         If an ArrayLike that cannot be cast is passed in.
 
     """
-    if np.can_cast(np.asarray(var_in), np.float64):
-        return np.asarray(var_in).astype(np.float64, casting="safe")
-    err_msg = f"Inappropriate input for handicaps code. Must be numeric value."
-    raise TypeError(err_msg)
+    try:
+        return np.asarray(var_in, dtype=np.float64)
+    except ValueError as exc:
+        err_msg = f"Inappropriate input for handicaps code. Must be numeric value."
+        raise TypeError(err_msg) from None
 
 
 class HandicapScheme(ABC):
@@ -730,5 +730,5 @@ class HandicapScheme(ABC):
 
         """
         val = self.score_for_round(hc_est, round_est, arw_d=arw_d, rounded_score=False)
-        # val is known to be a 0D array, so cast to float for type checker
-        return cast(float, val) - score_est
+        # val is known to be a 0D array, so cast to float for subsequent use
+        return float(val) - score_est
--- a/archeryutils/handicaps/handicap_scheme.py
+++ b/archeryutils/handicaps/handicap_scheme.py
@@ -34,7 +34,7 @@ References
 import itertools as itr
 import warnings
 from abc import ABC, abstractmethod
-from typing import TypeVar, cast
+from typing import cast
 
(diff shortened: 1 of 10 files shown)


jatkinson1000:
How far should `ArrayLike` be allowed to propagate in the code?

`ArrayLike` includes a numpy array or list of strings, and we should check that inputs can be cast to numbers before too deep. Probably using `np.can_cast()` https://numpy.org/doc/stable/reference/generated/numpy.can_cast.html as:
```python
if not numpy.can_cast(np.asarray(input), np.float64):
    Raise("Inappropriate input for code. Must be numeric value.")
```

Perhaps wrap this up as a reuseable function, and even use it to return the input cast to a `np.float64` array?

archeryutils/handicaps/handicap_scheme.py
```diff
@@ -715,4 +730,5 @@ def _f_root(
 
         """
         val = self.score_for_round(hc_est, round_est, arw_d=arw_d, rounded_score=False)
-        return val - score_est
+        # val is known to be a 0D array, so cast to float for type checker
+        return cast(float, val) - score_est
```
LiamPattinson, line 734: If `val` is a 0D array, `val - score_est` will be of type `np.float64`. Rather than using `cast`, if you instead do `float(val) - score_est`, the result will actually be of type `float`.

archeryutils/handicaps/handicap_tables.py
```diff
@@ -267,15 +267,13 @@ def _check_print_table_inputs(
...
-                msg = "Expected float or ndarray for hcs."
-                raise TypeError(msg)
-        hcs = hcs_in.astype(float)
+        hcs_in = np.asarray(hcs_in)  # Explicitly cast to array for calculations
+
+        try:
+            hcs = hcs_in.astype(np.float64)
+        except ValueError as exc:
+            msg = "Cannot convert supplied handicaps to float for HandicapTable."
+            raise TypeError(msg) from exc
```
LiamPattinson, line 276:
```suggestion
        try:
            hcs = np.asarray(hcs_in, dtype=float)
        except ValueError as exc:
            msg = "Cannot convert supplied handicaps to float for HandicapTable."
            raise TypeError(msg) from exc
```

archeryutils/handicaps/handicap_scheme.py
```diff
@@ -34,14 +34,37 @@
...
+    Raises
+    ------
+    TypeError
+        If an ArrayLike that cannot be cast is passed in.
+
+    """
+    if np.can_cast(np.asarray(var_in), np.float64):
+        return np.asarray(var_in).astype(np.float64, casting="safe")
+    err_msg = f"Inappropriate input for handicaps code. Must be numeric value."
+    raise TypeError(err_msg)
```
LiamPattinson, line 67: Rather than explicitly checking `can_cast`, I'd normally just use `np.asarray(..., dtype=float)` and do it all in one step. I think that could replace the use of the `_cast_float_array` function altogether, though it might be worth keeping around if you specifically need to raise a `TypeError` on failure (weirdly, `np.asarray(...)` raises a `ValueError` if the types can't be converted)

LiamPattinson (commented):
I think this all looks better now. Just have a few comments.

---

> How far should `ArrayLike` be allowed to propagate in the code?

I think the way you've done it is sensible. The rules I generally follow are:

- If you call any NumPy functions on some input, type it as `ArrayLike` and use `np.asarray(x)` at the top of the function.
- If you only pass `ArrayLike` things to other functions taking `ArrayLike`, simply propagate it forward.
- If something _must_ be a NumPy array on entry, instead type it as `npt.NDArray[np.float64]` (or something similar).

archeryutils/handicaps/handicap_scheme.py
jatkinson1000: Simplified in https://github.com/jatkinson1000/archeryutils/pull/133/commits/954a89bbba53c483db18b5cabd1eb91fac2ad010 but decided to keep the function to get the friendlier type error given it triggers quite a way down a call stack.

archeryutils/handicaps/handicap_tables.py
jatkinson1000: done in https://github.com/jatkinson1000/archeryutils/pull/133/commits/954a89bbba53c483db18b5cabd1eb91fac2ad010 but explicitly using `np.float64` to match the rest of the code.

archeryutils/handicaps/handicap_scheme.py
jatkinson1000: Good call! thanks.

PR #134, opened 2025-05-11
Add calculate_age_group

mjtamlyn:
This utility function works out the age group under AGB rules, given a year of birth and the year the event takes place in.

I've written this (or something like it) more than once, and it seems like a "core" utility function for AGB applications, usable in records code, entry systems, rankings etc.

I *think* I've done most of the docstrings, testing, type checks etc in roughly the right way, but let me know if htere's things missing.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   snapshots in [#148](https://github.com/jatkinson1000/archeryutils/pull/148)
 - Addition of strict_rounds and strict_distance arguments for classifications in
   [#149](https://github.com/jatkinson1000/archeryutils/pull/149)
+- Add function to calculate Archery GB age groups as part of the classifications module
+  by [@mjtamlyn](https://github.com/mjtamlyn) in [#134](https://github.com/jatkinson1000/archeryutils/pull/134)
 
 ### Changed
 
--- a/archeryutils/classifications/AGB_ages.json
+++ b/archeryutils/classifications/AGB_ages.json
@@ -2,6 +2,7 @@
   "OVER_50" : {
     "age_group" : "50+",
     "min_age": 50,
+    "max_age": null,
     "male" : [70, 80],
     "female" : [60, 60],
     "sighted" : [60, 60],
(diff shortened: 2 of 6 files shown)


mjtamlyn:
The failing test on python 3.10 relates to the use of `typing.NotRequired` for the optional keys `min_age` and `max_age`. It doesn't look like this has an obvious polyfill, and I don't think I know enough about python typing to know how to work around it.

jatkinson1000 (commented):
Thanks for this @mjtamlyn I think it's a really useful utility

Did some slight restructuring and added `null` to the json to satisfy Python 3.10 with all ages having a max/min which could be `None`. Also relocated the tests to follow the current structure.

Hopefully this all still works for the situations in which you envisage using this in your code, but let me know if not!

My only comment would be whether it is worth renaming the function to `calculate_agb_age_group()`? This would match the call signs for the agb classifications and leave open to implementing other schemes in future.
Edit: I'm currently implementing Archery Australia classifications - I wonder if passing an optional argument for rules (like with handicaps) might be useful. e.g. `"AGB"` vs `"AA"`.

mjtamlyn:
Thanks for the updates to make it work with the strict typing.

I'm happy for now to look at just renaming to `calculate_agb_age_group` as that's going to cover the majority of use cases for now. However, we should think about different "age group schemes" in more detail. Some use cases include:
- Archery Australia as you mentioned
- Strict WA age groups, which is technically a subset of AGB age groups
- Extended age groups, for example some events might run U10s, or the AGB 50+ event runs 65+
I agree I was only have way there with providing alternate JSON, so you'd also need to provide an alternate `enum`.

Even if we factor this out in the long term, it's reasonable to expose `calculate_agb_age_group` with the current API, and then to implement a more generic function in the future that this uses.

So I reckon just rename the fn and we're good to go?

jatkinson1000:
@mjtamlyn I just rebased, updated to use the new ages enum, and renamed the function.

I believe this is good to go if you can take a quick look and are happy.

My final thought is whether this should be in the classifications module or a new module (`rules`, `ages`, `governing_bodies`?), as it could be used for a whole bunch of things besides classifications?

Let me know your thoughts and approval and then I'll look to merge.

mjtamlyn:
PR in its current form looks great. I'm indifferent about its location.

PR #160, opened 2026-01-01
Organise json files, cache loading.

### LINKED ISSUE #68: Data from read_json functions could be stored as constants
I noticed during work on #67 that the read_json functions are called seperately in each AGB disciplines `make_classification_dict` function and stored to locals.

These could be called once and saved as constants, as any change to the data being loaded would require installing an updated version of archeryutils.

TomHall2020:
Small set of changes that would close #68 - does moving the json files cause any problems for your other systems @jatkinson1000 ? Was finding it tricky that they mixed in with the python files when sorted alphabetically.

Originally had thought of loading the data into global variables/constants - not sure if that actually helps anything compared to this approach of just caching the loading function - in all cases the behaviour is identical as the package loads and uses all the data at import time.

Reluctant to do any more on this as I feel the entire classification module structure needs an overhaul - too many modules with too many similar long names, but that is inevitable anyway for any refactoring to make a generic function ala #24 or #103 

Bonus extras in the second two commits if you want it. One to clean up reading of the classes files - since we're min version 3.10 now can use structural pattern matching syntax to replace the if-else statements for selecting the class files. And one to remove the unused filepath parameter from the age/bowstyle/gender files - its all used at import time, and we are assuming the contents of the json files are correct in both the code and the type hints so there isn't really much purpose to load an alternative file at runtime - thats already not an option for the indoor/outdoor classes files, and it simplifies the signature.


jatkinson1000 (commented):
Thanks @TomHall2020 

Happy with all the proposals here.
Agree this is neater, and the caching is a nice touch.

Happy with removing the optional file in the call signatures - this was in the old code and as you say it has evolved to a point where no other file would be passed in. I think it is nice retaining the data as a json file as it's easy to read and edit.

Any reason to use the `json.loads(file.read_text())` over `json.load(file)` on an open file context?

Happy to overlook the failing coverage patch as those lines were not covered in the previous code.

Please could you drop something in the changelog to summarise.

